PR #1, opened 2018-06-05
Add tests for the consumer

abompard:
I'm not pushing this right now because I think we need to have a discussion on what we want to do when a message can't be decoded or isn't JSON. Currently we just raise `ValidationError` which, I believe, crashes the consumer. Do we want to do a `basic_nack()` too? Do we want to log but ignore the message?

### DIFF
--- a/fedora_messaging/_session.py
+++ b/fedora_messaging/_session.py
@@ -392,6 +392,11 @@ class ConsumerSession(object):
 
 
 def get_message(routing_key, properties, body):
+    if properties.headers is None:
+        _log.error('Message (body=%r) arrived without headers. '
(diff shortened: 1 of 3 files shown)


jeremycline:
Hmm, yes. We could nack with ``requeue=False`` so we have the option of sending the message to a dead-letter exchange (we don't currently set those up, but perhaps we should?). It'd also be good to log at the error level so it gets attention, then carry on with the next message. That feels like a nice compromise of user-friendliness without encouraging invalid message formats to me.

abompard:
OK, I made those changes. The error logging was already setup.

PR #3, opened 2018-06-15
Add settings for TLS client authentication

jeremycline:
This is annoyingly complicated because the way to set things up changed with pika 1.0.0 (not yet released). This supports both. Once 1.0.0 is released and available for us to use, we can remove the old approach.


fedora_messaging/_session.py
```diff
@@ -21,25 +21,88 @@
...
+            'cert_reqs': ssl.CERT_REQUIRED,
+            'ssl_version': ssl.PROTOCOL_TLSv1_2,
+        }
+    else:
+        ssl_context = ssl.create_default_context()
+        try:
+            ssl_context.load_verify_locations(cafile=config.conf['tls']['ca_cert'])
+        except ssl.SSLError as e:
+            raise ConfigurationException(
+                'The "ca_cert" setting in the "tls" section is invalid: ' + str(e))
```
abompard: Using `+` to concatenate strings is... unusual :-) People use the `.format()` string method these days ;-)

fedora_messaging/tests/unit/test_session.py
```diff
@@ -41,22 +44,31 @@ def setUp(self):
...
         self.assertEqual(publisher._parameters.host, "localhost")
         self.assertEqual(publisher._parameters.port, 5672)
         self.assertEqual(publisher._parameters.virtual_host, "/")
-        self.assertEqual(publisher._parameters.ssl, False)
-        # Now test with a custom URL
-        publisher = _session.PublisherSession(
-            "amqps://username:password@rabbit.example.com/vhost",
-            "test_exchange",
-        )
+        self.assertTrue(publisher._parameters.ssl_options is None)
```
abompard: I think `assertIsNone()` is available.

fedora_messaging/tests/unit/test_session.py
```diff
@@ -41,22 +44,31 @@ def setUp(self):
...
+        with mock.patch.dict(config.conf, {'tls': self.tls_conf}):
+            publisher = _session.PublisherSession(
+                "amqps://username:password@rabbit.example.com/vhost",
+                "test_exchange",
+            )
         self.assertEqual(publisher._parameters.host, "rabbit.example.com")
         self.assertEqual(publisher._parameters.port, 5671)
         self.assertEqual(publisher._parameters.virtual_host, "vhost")
-        self.assertEqual(publisher._parameters.ssl, True)
+        self.assertTrue(publisher._parameters.ssl_options is not None)
```
abompard: I think `assertIsNotNone()` is available.

abompard:
I made a couple minor comments, but it does look very good to me, thanks! Merge when you want.

fedora_messaging/_session.py
jeremycline: I figured since it was at the end of the string it was shorter (LOC-wise) to just concatenate, but I'm happy to switch it to format

abompard: As you wish, but `.format()` opens a parenthesis which lets you indent nicely on the new line ;-)

PR #5, opened 2018-07-12
Modify check for pika TLS API

jeremycline:
0.12 adds the SSLOptions object, but it's got a different API than
1.0.0, and the old dictionary assignment still works. Just mark
SSLOptions as None if it's version 0.12 (this will break if there's a
0.13, but I don't think there will be).


abompard:
LGTM, feel free to merge :-)

PR #7, opened 2018-07-18
Twisted support

abompard:
This branch adds Twisted support using Pika's Twisted adapter. Pika's Twisted integration is currently being heavily worked on so there are version checks to maintain compatibility with the current released version.


fedora_messaging/twisted/protocol.py
```diff
@@ -0,0 +1,296 @@
...
+class FedoraMessagingProtocol(TwistedProtocolConnection):
+    """A Twisted Protocol for the Fedora Messaging system.
+
+    This protocol builds on the generic pika AMQP protocol to add calls
+    specific to the Fedora Messaging implementation.
+    """
+
+    name = 'FedoraMessaging:Protocol'
+
+    def __init__(self, parameters, confirms=False):
```
jeremycline: Since this isn't exposed as a public API it shouldn't matter, but it seems like confirms=True is best since it's safer and probably the option you want unless you know what you're doing.

fedora_messaging/twisted/protocol.py
```diff
@@ -0,0 +1,296 @@
...
+    @defer.inlineCallbacks
+    def connectionReady(self, res=None):
+        """Called when the AMQP connection is ready.
+        """
+        # The optional `res` argument is for compatibility with pika < 1.0.0
+        # Create channel
+        self._channel = yield self.channel()
+        log.msg("AMQP channel created", system=self.name,
+                logLevel=logging.DEBUG)
+        yield self._channel.basic_qos(prefetch_count=0, prefetch_size=0)
```
jeremycline: There's a setting for QoS: https://github.com/fedora-infra/fedora-messaging/blob/fd4d4cc6a9ec75cc5af090fd81f9fc21f2d30d05/fedora_messaging/config.py#L208

fedora_messaging/twisted/protocol.py
```diff
@@ -0,0 +1,296 @@
...
+        on the user-provided callback, see the user guide on consuming.
+
+        Args:
+            delivery_frame (pika.spec.Deliver): The delivery frame which
+                includes details about the message like content encoding and
+                its delivery tag.
+            properties (pika.spec.BasicProperties): The message properties like
+                the message headers.
+            body (bytes): The message payload.
+            message_callback (callable): the function that will be called with
```
jeremycline: This looks to no longer be a method argument and is instead set in the `setupRead` call.

fedora_messaging/twisted/service.py
```diff
@@ -0,0 +1,97 @@
...
+    name = 'fedora-messaging'
+    factoryClass = FedoraMessagingFactory
+
+    def __init__(self, on_message, amqp_url=None, bindings=None):
+        """Initialize the service.
+
+        Args:
+            on_message (callable|None): Callback that will be passed each
+                incoming messages. If None, no message consuming is setup.
+            amqp_url (str): URL to use for the AMQP server.
```
jeremycline: ``__init__`` also accepts ``bindings``, looks like it's the same format documented in config

fedora_messaging/twisted/service.py
```diff
@@ -0,0 +1,97 @@
...
+        return None
+
+    def connect(self):
+        factory = self.factoryClass(self._parameters, self._bindings)
+        if self._parameters.ssl_options:
+            serv = SSLClient(
+                host=self._parameters.host,
+                port=self._parameters.port,
+                factory=factory,
+                contextFactory=ssl.ClientContextFactory(),
```
jeremycline: Do we need to configure the client context ourselves for client x509 certificates, SSL/TLS versions, etc? 

 It looks like it just has one method, ``getContext()``, that needs to return a PyOpenSSL [Context](https://pyopenssl.readthedocs.io/en/stable/api/ssl.html#OpenSSL.SSL.Context) instance.

fedora_messaging/twisted/factory.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+        yield self._client_ready
+        try:
+            yield self.client.publish(message, exchange)
+        except (
+                pika.exceptions.ConnectionClosed, pika.exceptions.ChannelClosed
+                ) as e:
+            log.msg('Connection lost while publishing, retrying.',
+                    system=self.name, logLevel=logging.WARNING)
+            yield self.publish(message, exchange)
```
jeremycline: Will this eventually time out? I guess we do want it to try until it succeeds or the service is shut down.

fedora_messaging/twisted/protocol.py
abompard: Yeah, I set it to False by default because it wasn't implemented in pika's Twisted integration anyway. My pika PR now contains it, but it's still unreleased. Also it's a RabbitMQ extension IIRC.

fedora_messaging/twisted/service.py
abompard: Actually, I haven't tested SSL/TLS connections with client certificate yet... I don't know.

fedora_messaging/twisted/factory.py
abompard: Hmm I don't think it'll time out, it should just wait for the new connection to be ready and publish afterwards (waiting for the `_client_ready` Deferred). Did I understand your question correctly?

abompard:
Yeah I should write tests now, if you think it's looking good enough :-)

fedora_messaging/twisted/factory.py
jeremycline: Yeah, that was what I thought. It may be that someday someone will want to publish something that errbacks after a timeout is reached rather than retrying forever, but that's not something we need to add right now.

abompard:
Alright, all tests pass now.

fedora_messaging/twisted/protocol.py
```diff
@@ -0,0 +1,297 @@
...
+        log.msg("AMQP channel created", system=self.name,
+                logLevel=logging.DEBUG)
+        yield self._channel.basic_qos(
+            prefetch_count=config.conf["qos"]["prefetch_count"],
+            prefetch_size=config.conf["qos"]["prefetch_size"],
+        )
+        if self._confirms:
+            yield self._channel.confirm_delivery()
+        if _pika_version < pkg_resources.parse_version("1.0.0"):
+            TwistedProtocolConnection.connectionReady(self, res)
```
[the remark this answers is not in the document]
abompard: Yeah, on Python2 the parent `Protocol` class does not inherit from `object`, it's an "old-style class" (yes, they still exist apparently), so `super()` does not work. Die, Python2.

fedora_messaging/twisted/protocol.py
```diff
@@ -0,0 +1,297 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+from __future__ import absolute_import, unicode_literals
```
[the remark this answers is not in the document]
abompard: Actually I'll do the opposite, keep `unicode_literals` and mark strings passed to logging with `b""`.

fedora_messaging/twisted/protocol.py
```diff
@@ -0,0 +1,297 @@
...
+        """Pass incoming messages to the provided callback.
+
+        Args:
+            message_callback (callable): The callable to pass the message to
+                when one arrives.
+        """
+        if not self.factory.bindings:
+            return
+        self._message_callback = message_callback
+        for binding in self.factory.bindings:
```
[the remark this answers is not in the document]
abompard: I fixed the doc. What about this config:
```
[[bindings]]
queue = "my_queue_1"
exchange = "amq.topic"
routing_keys = ["key1"]

[[bindings]]
queue = "my_queue_2"
exchange = "amq.topic"
routing_keys = ["key2"]
```
Or is using the double-bracket feature of TOML a bit too obscure? I think it's fine.

fedora_messaging/twisted/protocol.py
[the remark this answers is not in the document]
jeremycline: From the sounds of the documentation, it's an error to pass unicode strings in Python 2 and an error to pass bytes in Python 3, but that's just my interpretation.

fedora_messaging/twisted/protocol.py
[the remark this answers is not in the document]
jeremycline: Ooooof

fedora_messaging/twisted/protocol.py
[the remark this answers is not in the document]
jeremycline: I think it's fine, and there's [this](https://github.com/fedora-infra/fedora-messaging/blob/4303b66eadf334456d2ef7b202bf074cb57bdb16/config.toml.example#L10) alternate syntax to make it easier.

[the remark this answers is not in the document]
abompard: Yeah I think the double-bracket syntax is less error-prone.

fedora_messaging/twisted/protocol.py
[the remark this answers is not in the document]
abompard: Meh. I've had hard-to-debug issues in the past when not using `unicode_literals`. But if that's the only way to go, I'll change it.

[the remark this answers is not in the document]
jeremycline: Yeah, it's a bit of a toss-up (and also a bit weird Twisted won't accept unicode in Py2).

Die, Python 2, die.

ktdreyer:
Really cool to see Twisted support here.

PR #8, opened 2018-07-18
Don't hang on Python 2.7 if the broker isn't running

jeremycline:
It looks like Python 3 will kill the process a connection error is
raised, but Python 2 doesn't. This adds some explicit cleanup.


abompard:
LGTM!

PR #9, opened 2018-07-23
Add a sample message schema Python package with documentation

jeremycline:

### DIFF
--- a/docs/messages.rst
+++ b/docs/messages.rst
@@ -47,75 +47,55 @@ The default body schema simply declares that the header field must be a JSON
 object.
 
 
-Example
-=======
+Example Schema
+--------------
 
-An example is worth a thousand words::
-
-    from fedora_messaging import message
-
-    class MailmanMessage(message.Message):
-        """
-        A sub-class of a Fedora message that defines a message schema for
-        messages published by Mailman when it receives mail to send out.
-        """
-        body_schema = {
-            'id': 'http://fedoraproject.org/message-schema/mailman#',
-            '$schema': 'http://json-schema.org/draft-04/schema#',
-            'description': 'Schema for message sent to mailman',
-            'type': 'object',
-            'properties': {
-                'mlist': {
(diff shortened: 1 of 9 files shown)


abompard:
It looks good in itself, but I'm wondering if we shouldn't separate the schema from the `Message` class to make sure devs don't do weird things by subclassing the `validate()` method. I've tried this split here[1], what do you think?

[1] https://github.com/abompard/fedora-messaging/commit/041b4689face536ee6c1ba76b25dad98c0eb8217

I've also separated the methods that extract (human-readable) information, because those methods don't have to change with each schema revision, so I though we could point to the message "extractor" or "parser" class. (I don't particularly like the term "extractor").

jeremycline:
> It looks good in itself, but I'm wondering if we shouldn't separate the schema from the Message class to make sure devs don't do weird things by subclassing the validate() method. I've tried this split here[1], what do you think?
>
> I've also separated the methods that extract (human-readable) information, because those methods don't have to change with each schema revision, so I though we could point to the message "extractor" or "parser" class. (I don't particularly like the term "extractor").

I think we should make an effort to stop developers from accidentally shooting themselves in the foot, but I don't think we should (or even can) stop developers from that kind of failing. We should document that ``validate`` should not be overridden in a sub-class if the developer wants any benefit from the schema they just wrote, of course, but a truly determined individual can overcome whatever obstacles we put in their way if they want to make their lives difficult.

I do think the separation of "parser" from schema is interesting, but I don't think we need to enforce that pattern. A user could write their own ``MessageExtractor`` class and use multiple inheritance with it and the ``Message`` class rather than having it as attributes. Or, if they don't like that, they can write the one class. I'm generally nervous about adding lots of tiny classes that need to be combined to make a useful object. Might be worth documenting that approach, though?

abompard:
Yeah that makes sense, I'll remove that commit. This PR can be merged then.

PR #11, opened 2018-07-24
Increase test coverage by removing dead code and testing config

jeremycline:
Remove the now-unused constants module and test config.py.


fedora_messaging/config.py
```diff
@@ -393,9 +356,56 @@ def setup_logging(self):
...
+        Perform checks on the configuration to assert its validity
+
+        Raises:
+            ConfigurationException: If the configuration is invalid.
+        """
+        for key in self:
+            if key not in DEFAULTS:
+                raise exceptions.ConfigurationException(
+                    'Unknown configuration key "{}"! Valid configuration keys are'
+                    ' {}'.format(key, list(DEFAULTS.keys())))
```
bowlofeggs: I would love to be able to do this in Bodhi's config too, but Bodhi has *dynamic* config names. Ugh.

fedora_messaging/config.py
```diff
@@ -393,9 +356,56 @@ def setup_logging(self):
...
+
+        # Don't let users override these
+        self['client_properties']['version'] = DEFAULTS['client_properties']['version']
+        self['client_properties']['information'] = DEFAULTS['client_properties']['information']
+        self['client_properties']['product'] = DEFAULTS['client_properties']['product']
```
bowlofeggs: It might be good to check if these are overridden and raise an exception. This way users will know that their overridden settings aren't accepted.

fedora_messaging/config.py
```diff
@@ -393,9 +356,56 @@ def setup_logging(self):
...
+        else:
+            config_path = '/etc/fedora-messaging/config.toml'
+
+        if os.path.exists(config_path):
+            _log.info('Loading configuration from {}'.format(config_path))
+            with open(config_path) as fd:
+                try:
+                    file_config = pytoml.loads(fd.read())
+                    for key in file_config:
+                        config[key.lower()] = file_config[key]
```
bowlofeggs: It might be good to be opinionated about case, but I don't feel strongly about it.

fedora_messaging/tests/unit/test_config.py
```diff
@@ -0,0 +1,301 @@
...
+        """Assert loading the config with an empty file that exists works."""
+        config = msg_config.LazyConfig()
+        self.assertEqual(msg_config.DEFAULTS['callback'], config.get('callback'))
+        self.assertEqual(msg_config.DEFAULTS['amqp_url'], config.get('amqp_url'))
+        mock_exists.assert_called_once_with('/etc/fedora-messaging/config.toml')
+        mock_log.info.assert_called_once_with(
+            'Loading configuration from /etc/fedora-messaging/config.toml')
+
+    def test_explode_on_pop(self):
+        """Assert loading the config with an empty file that exists works."""
```
bowlofeggs: This description seems wrong for the test.

fedora_messaging/tests/unit/test_config.py
```diff
@@ -0,0 +1,301 @@
...
+        self.assertEqual(msg_config.DEFAULTS, copy)
+        mock_exists.assert_called_once_with('/etc/fedora-messaging/config.toml')
+        mock_log.info.assert_called_once_with(
+            'Loading configuration from /etc/fedora-messaging/config.toml')
+
+    @mock.patch('fedora_messaging.config.open', mock.mock_open(read_data=empty_config))
+    @mock.patch('fedora_messaging.config._log', autospec=True)
+    @mock.patch('fedora_messaging.config.os.path.exists', return_value=True)
+    def test_load_on_get(self, mock_exists, mock_log):
+        """Assert loading the config with an empty file that exists works."""
```
bowlofeggs: This comment and a few above seem to describe a different test.

fedora_messaging/config.py
jeremycline: @abompard, any opinion here? I don't have strong feelings about it, either.

abompard: Me neither. I don't imagine we'll have keys that will only differ by case.

PR #16, opened 2018-08-01
Have the Message class carry the properties as an attribute

abompard:
I wonder if you think that would be a good idea. The Message class is already carrying the topic, I think it makes sense if the properties are stored there too.

One big plus: the `message_id` would be created on instantiation and would be always accessible, which is pretty cool to log when something goes wrong or to identify a message.

What do you think @jeremycline ?

### DIFF
--- a/fedora_messaging/_session.py
+++ b/fedora_messaging/_session.py
@@ -20,13 +20,11 @@ import inspect
 import json
 import logging
 import signal
-import uuid
 import ssl
 
 from pika import exceptions as pika_errs
 import pika
 import pkg_resources
-import jsonschema
 try:
     # Versions of pika greater than the 0.11 series uses SSLOptions to configure
     # a TLS connection.
@@ -35,7 +33,7 @@ except ImportError:
     SSLOptions = None
 
 from . import config
-from .message import _schema_name, get_class, Message
+from .message import get_message
 from .exceptions import (
     Nack, Drop, HaltConsumer, ValidationError, PublishReturned, ConnectionException,
     ConfigurationException)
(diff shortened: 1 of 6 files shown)


fedora_messaging/message.py
```diff
@@ -85,6 +89,46 @@ def _schema_name(cls):
     return '{}:{}'.format(cls.__module__, cls.__name__)
 
 
+def get_message(routing_key, properties, body):
```
jeremycline: A doc-block here would be good

fedora_messaging/message.py
```diff
@@ -127,11 +171,37 @@ class Message(object):
         'type': 'object',
     }
 
-    def __init__(self, body=None, headers=None, topic=None):
-        self.headers = headers or {}
+    def __init__(self, body=None, headers=None, topic=None, properties=None):
```
jeremycline: Please add "properties" to the docblock

fedora_messaging/_session.py
```diff
@@ -160,14 +157,17 @@ def publish(self, message, exchange=None):
...
-    def _connect_and_publish(self, exchange, routing_key, body, properties):
+    def _connect_and_publish(self, exchange, message):
         if not self._connection or not self._channel:
             self._connection = pika.BlockingConnection(self._parameters)
             self._channel = self._connection.channel()
             if self._confirms:
                 self._channel.confirm_delivery()
 
-        self._channel.publish(exchange, routing_key, body, properties)
+        body = json.dumps(message.body).encode('utf-8')
```
jeremycline: If you wanted, you could add the body and routing key as properties (called 'encoded_body' and 'encoded_routing_key' or something) so you don't have to Just Know the body needs to be utf-8 encoded.

Maybe even just add it as a private property ('_encoded_body', etc) since I can't think of a reason for anything outside this library needing this.

PR #19, opened 2018-08-03
Add tutorial

abompard:
This branch adds a tutorial that we'll be able to use in the Flock workshop.

### DIFF
--- a/docs/tutorial/conversion.rst
+++ b/docs/tutorial/conversion.rst
@@ -1,0 +1,292 @@
+Converting a fedmsg application
+===============================
+
+.. highlight:: python
+
+Converting a Flask app
+----------------------
+
+.. elections, fedocal
+
+Let's use the `elections`_ app as an example. Clone the code using the following command::
+
+    git clone https://pagure.io/elections.git
+
+And change to this directory.
+
+.. _elections: https://pagure.io/elections/
+
+In the ``elections`` app, all calls to publish messages on fedmsg are going
+through the ``fedora_elections.fedmsgshim.publish`` wrapper function. We can
+thus modify this function to make it call Fedora Messaging instead of fedmsg.
+
+JSON schema
+~~~~~~~~~~~
+First, you will need a Message schema. To write this schema you must know what
+kind of messages are sent on the bus. A ``git grep`` command will reveal that
+all calls are made from the ``admin.py`` file. Open that file and examine those
+calls.
+
+In parallel, copy the ``docs/sample_schema_package/`` directory from the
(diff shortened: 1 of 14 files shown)


docs/tutorial/conversion.rst
```diff
@@ -0,0 +1,292 @@
+Converting a fedmsg application
```
jeremycline: I think in the other docs the header convention has been:
```
=========================
over and underlined for top level
=========================

double underlined for next level
=========================

single underlined for 3rd level
------------------------------------------

get inventive from here on out
~~~~~~~~~~~~~~~~~~~~~~~~

```

docs/tutorial/conversion.rst
```diff
@@ -0,0 +1,292 @@
...
+===============================
+
+.. highlight:: python
+
+Converting a Flask app
```
jeremycline: Maybe put this inside a "Converting Publishers" section for clarity

docs/tutorial/conversion.rst
```diff
@@ -0,0 +1,292 @@
...
+    [[bindings]]
+    queue = "tutorial"
+    exchange = "amq.topic"
+    routing_keys = ["fedora.elections.#"]
+
+You could also add ``"fedora.elections.#"`` to the ``"routing_keys"`` value in
+the existing entry.
+
+Now make sure that RabbitMQ is still running, and run the ``consume.py`` script we
+used before. Make sure it is not systematically raising exceptions in the
```
jeremycline: It can be helpful to link to the usage.rst document here where the consume.py script is first mentioned, just in case the reader forgot where they saw it

docs/tutorial/exceptions.rst
```diff
@@ -0,0 +1,54 @@
+Handling exceptions
+===================
+
+.. highlight:: python
+
+All exceptions are located in the ``fedora_messaging.exceptions`` module.
+
+When publishing
```
jeremycline: It'd be good if all the information here was neatly included in the API documentation. I see I didn't document that ``publish`` raises a ValidationError, so if we included that and more detailed doc-blocks on the exception classes, you should be able to just point to the ``fedora_messaging.api.publish`` API docs.

abompard: I fixed the rest but I'm not entirely sure how you picture that part. I've added the references and enhanced the doc-blocks. Do you mean this entire section should be replaced by a reference to the API docs?

jeremycline: Just more helpful error handling documentation for the publish API. I can shuffle things around in a follow-up PR so feel free to merge this.

PR #21, opened 2018-08-03
Add tests for the CLI

jeremycline:
Also fix a small bug from exception changes in config.py

### DIFF
--- a/fedora_messaging/cli.py
+++ b/fedora_messaging/cli.py
@@ -24,6 +24,7 @@ from __future__ import absolute_import
 import importlib
 import logging
 import logging.config
+import os
 import sys
 
 import click
@@ -72,10 +73,12 @@ _amqp_url_help = (
 def cli(conf):
     """The fedora-messaging command line interface."""
     if conf:
+        if not os.path.isfile(conf):
+            raise click.exceptions.BadParameter('{} is not a file'.format(conf))
         try:
             config.conf.load_config(config_path=conf)
-        except ValueError as e:
-            raise click.exceptions.BadParameter(e)
+        except exceptions.ConfigurationException as e:
+            raise click.exceptions.BadParameter(str(e))
         config.conf.setup_logging()
 
 
@@ -87,7 +90,7 @@ def cli(conf):
 @click.option('--exchange', help=_exchange_help)
 @click.option('--amqp-url', help=_amqp_url_help)
 def consume(amqp_url, exchange, queue_name, routing_key, callback, app_name):
-
+    """Consume messages from an AMQP queue using a Python callback."""
     amqp_url = amqp_url or config.conf['amqp_url']
     if exchange and queue_name and routing_key:
         bindings = [{
--- a/fedora_messaging/config.py
+++ b/fedora_messaging/config.py
@@ -404,8 +404,9 @@ class LazyConfig(dict):
                     for key in file_config:
                         config[key.lower()] = file_config[key]
                 except pytoml.core.TomlError as e:
-                    _log.error('Failed to parse {}: {}'.format(config_path, str(e)))
-                    raise exceptions.ConfigurationException(e)
+                    msg = 'Failed to parse {}: error at line {}, column {}'.format(
+                            config_path, e.line, e.col)
+                    raise exceptions.ConfigurationException(msg)
         else:
             _log.info('The configuration file, {}, does not exist.'.format(config_path))
 
--- a/fedora_messaging/exceptions.py
+++ b/fedora_messaging/exceptions.py
@@ -6,7 +6,19 @@ class BaseException(Exception):
 
 
 class ConfigurationException(BaseException):
-    """Raised when there's an invalid configuration setting"""
+    """
+    Raised when there's an invalid configuration setting
+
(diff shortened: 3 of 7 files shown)


fedora_messaging/cli.py
```diff
@@ -87,7 +90,7 @@ def cli(conf):
 @click.option('--exchange', help=_exchange_help)
 @click.option('--amqp-url', help=_amqp_url_help)
 def consume(amqp_url, exchange, queue_name, routing_key, callback, app_name):
-
+    """Consume messages from an AMQP queue using a Python callback."""
```
bowlofeggs: Might be nice to also document the params and their types, but this is still an improvement over what was there before so feel free to defer.

fedora_messaging/config.py
```diff
@@ -404,8 +404,9 @@ def load_config(self, config_path=None):
                     for key in file_config:
                         config[key.lower()] = file_config[key]
                 except pytoml.core.TomlError as e:
-                    _log.error('Failed to parse {}: {}'.format(config_path, str(e)))
-                    raise exceptions.ConfigurationException(e)
+                    msg = 'Failed to parse {}: error at line {}, column {}'.format(
+                            config_path, e.line, e.col)
```
bowlofeggs: Nice.

fedora_messaging/exceptions.py
```diff
@@ -8,6 +8,12 @@ class BaseException(Exception):
 class ConfigurationException(BaseException):
     """Raised when there's an invalid configuration setting"""
 
+    def __init__(self, message):
+        self.message = message
```
bowlofeggs: Might be good to document the type of ```message```.

fedora_messaging/tests/fixtures/bad_conf.toml
```diff
@@ -0,0 +1 @@
+publish_exchange = I forgot quotes here
```
bowlofeggs: heh, GitHub does not like this.

fedora_messaging/tests/unit/test_cli.py
```diff
@@ -0,0 +1,83 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""Tests for the :module:`fedora_messaging.cli` module."""
+
+import os
+import unittest
+
+from click.testing import CliRunner
+import mock
```
bowlofeggs: In Python 3 mock is a built-in, and we're *almost* Python3-y.

fedora_messaging/tests/unit/test_cli.py
```diff
@@ -0,0 +1,83 @@
...
+                echo,
+                [{'exchange': 'e', 'queue_name': 'q', 'routing_key': '#'}])
+        self.assertEqual(0, result.exit_code)
+
+    @mock.patch('fedora_messaging.cli.api.consume')
+    def test_bad_conf(self, mock_consume):
+        """Assert a bad configuration file is reported."""
+        runner = CliRunner()
+        result = runner.invoke(cli.cli, ['--conf=' + BAD_CONF, 'consume'])
+        self.assertEqual(2, result.exit_code)
```
bowlofeggs: It might be good to check that the error message contains some expected text.

fedora_messaging/tests/unit/test_cli.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @mock.patch('fedora_messaging.cli.api.consume')
+    def test_missing_conf(self, mock_consume):
+        """Assert a missing configuration file is reported."""
+        runner = CliRunner()
+        result = runner.invoke(cli.cli, ['--conf=thispathdoesnotexist', 'consume'])
+        self.assertEqual(2, result.exit_code)
```
bowlofeggs: Similar here.

fedora_messaging/cli.py
jeremycline: I need to double check, but I believe click uses this docblock for user help, and it also renders the options help, so unfortunately it looks weird when you add the standard docblocks. I think I can add type hints to the click options, though.

fedora_messaging/tests/unit/test_cli.py
jeremycline: It makes me sad every time :crying_cat_face:

bowlofeggs:
On 08/17/2018 06:36 AM, Jeremy Cline wrote:
> I need to double check, but I believe click uses this docblock for user
> help, and it also renders the options help, so unfortunately it looks
> weird when you add the standard docblocks. I think I can add type hints
> to the click options, though.

Ah yes, this is true. Clement came up with a workaround that works,
though is kinda unconventional, in Bodhi's CLI. He basically put two
docblocks on the function. The first one gets used by click to render
help text, and the second gets using by nothing but humans, with their
soft human brains. Here's an example:

https://github.com/fedora-infra/bodhi/blob/3.8.1/bodhi/client/__init__.py#L208-L221

jeremycline:
Ah, neato.

PR #22, opened 2018-08-07
Hide the JSON body and properties from users of messages

jeremycline:
Publishers shouldn't need to touch these after creation, and consumers
should not be digging in the raw JSON as this makes migrating from one
schema to another difficult. Instead, we should wrap the JSON object in
Python methods which form the public API of the message.

### DIFF
--- a/docs/sample_schema_package/mailman_schema/schema.py
+++ b/docs/sample_schema_package/mailman_schema/schema.py
@@ -13,15 +13,44 @@
 # You should have received a copy of the GNU General Public License along
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""This is an example of a message schema."""
+
 from fedora_messaging import message
 
 
-class Message(message.Message):
+class BaseMessage(message.Message):
+    """
+    You should create a super class that each schema version inherits from.
+    This lets consumers perform ``isinstance(msg, BaseMessage)`` if they are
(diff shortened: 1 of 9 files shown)


docs/sample_schema_package/mailman_schema/schema.py
```diff
@@ -13,15 +13,44 @@
...
+    """
+    You should create a super class that each schema version inherits from.
+    This lets consumers perform ``isinstance(msg, BaseMessage)`` if they are
+    receiving multiple message types and allows the publisher to change the
+    schema as long as they preserve the Python API.
+    """
+
+    def __str__(self):
+        """Return a complete human-readable representation of the message."""
+        return 'Subject: {subj}\n\n{body}\n'.format(
```
abompard: Not sure about the empty line between the subject and the body, it makes for a very verbose output.

fedora_messaging/message.py
```diff
@@ -210,28 +205,30 @@ def _build_properties(self, headers):
         )
 
     @property
-    def headers(self):
-        return self.properties.headers
+    def _headers(self):
```
abompard: Maybe add a docstring stating that we would consider making this property public if there is a use case for that. Otherwise I'm not sure people will ask.

docs/sample_schema_package/mailman_schema/schema.py
jeremycline: Fixed!

fedora_messaging/message.py
jeremycline: Done, good idea

PR #26, opened 2018-08-15
Packaging

abompard:
Add and RPM spec file. I still get some rpmlint messages but I think they can be ignored: they're about files bundled in the testsuite.


fedora-messaging.spec
```diff
@@ -0,0 +1,132 @@
+%{!?__python2: %global __python2 %__python}
+%{!?python2_sitelib: %global python2_sitelib %(%{__python2} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")}
+
+%global pkgname fedora-messaging
+%global srcname fedora_messaging
+%global desc \
+Tools and APIs to make working with AMQP in Fedora easier.
```
bowlofeggs: IMO, spec files aren't that useful to have upstream since they are very distro specific, and it's painful to keep upstream and downstream synchronized. Bodhi used to have one in its GitHub repo and I removed it in favor of maintaining the one in dist-git.

setup.py
```diff
@@ -66,8 +66,7 @@ def get_requirements(requirements_file='requirements.txt'):
     maintainer_email='infrastructure@lists.fedoraproject.org',
     platforms=['Fedora', 'GNU/Linux'],
     keywords='fedora',
-    packages=find_packages(exclude=('fedora_messaging.tests',
-                                    'fedora_messaging.tests.*')),
+    packages=find_packages(),
```
bowlofeggs: Do you really want to ship the tests with the package? I think it's nice to exclude them from the production package.

abompard: I like the idea of being able to run the tests on the system you run the code on.

fedora-messaging.spec
abompard: Yeah I agree, I'm putting it here for review mainly. Then I'll probably remove it.

fedora-messaging.spec
```diff
@@ -0,0 +1,132 @@
+%{!?__python2: %global __python2 %__python}
+%{!?python2_sitelib: %global python2_sitelib %(%{__python2} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")}
```
jeremycline: Does EPEL7 not define these? I think both these macros should be defined everywhere we care about building this.

fedora-messaging.spec
jeremycline: I agree with @bowlofeggs. If you want to just take this specfile and submit it as a package to Fedora I'm happy to review it now. We can build it in rawhide only until we release 1.0.0, which will be long before f30 branches.

fedora-messaging.spec
```diff
@@ -0,0 +1,132 @@
...
+BuildRequires:  python3-mock
+BuildRequires:  python3-sphinx
+
+%description %{desc}
+
+%package     -n python2-%{pkgname}
+Summary:        %{summary}
+Requires:       %{name} = %{version}
+%{?python_provide:%python_provide python2-%{pkgname}}
+Requires:       python2-blinker
```
jeremycline: You can use the fancy new %{?python_enable_dependency_generator} macro to auto-generate the requires from the Python package requirements

fedora-messaging.spec
```diff
@@ -0,0 +1,132 @@
...
+
+
+%build
+%py2_build
+%py3_build
+# generate docs
+PYTHONPATH=${PWD} sphinx-build-3 -M html -d docs/_build/doctrees docs docs/_build/html
+PYTHONPATH=${PWD} sphinx-build-3 -M man -d docs/_build/doctrees docs docs/_build/man
+# remove the sphinx-build leftovers
+#rm -rf docs/_build/.{doctrees,buildinfo}
```
jeremycline: Leftover from something?

fedora-messaging.spec
```diff
@@ -0,0 +1,132 @@
...
+BuildRequires:  python2-blinker
+BuildRequires:  python2-click
+BuildRequires:  python2-jsonschema
+BuildRequires:  python2-pytoml
+BuildRequires:  python2-pika
+BuildRequires:  python2-six
+BuildRequires:  python2-pytest
+BuildRequires:  python2-twisted
+BuildRequires:  python2-pyOpenSSL
+#BuildRequires:  python2-pytest-twisted
```
jeremycline: It this not packaged or do we not need it?

fedora-messaging.spec
```diff
@@ -0,0 +1,132 @@
...
+
+
+Name:           %{pkgname}
+Version:        1.0.0
+Release:        0.1.a1%{?dist}
+Summary:        Set of tools for using Fedora's messaging infrastructure
+
+License:        GPLv2+
+URL:            https://github.com/fedora-infra/fedora-messaging
+Source0:        https://github.com/fedora-infra/%{name}/archive/v%{version}a1/%{srcname}-%{version}a1.tar.gz
```
jeremycline: You can also just do %{url}/archive... here if you want. It doesn't matter to me either way.

fedora-messaging.spec
abompard: oops :-)

fedora-messaging.spec
abompard: It's not packaged and we only need it for the Twisted unit tests. So I disabled those tests.

fedora-messaging.spec
abompard: nifty! :-)

fedora-messaging.spec
abompard: Yeah. And this currently does not build on EPEL7 because of missing dependencies (OpenSSL, blinker, and Twisted for Python 3).
Why do we need PyOpenSSL exactly?

jeremycline: I believe Twisted still relies on it for... Something. Possibly certificate chain validation? AFAIK cryptography still hasn't got that API.

PR #29, opened 2018-08-17
Fix authentication with x509 certificates

jeremycline:
The connection parameters need to set the credentials type to 'EXTERNAL'
in order for x509 client certificate authentication to work. Do so if
both the cert and key is present.


jeremycline:
Yes, I've got a branch in progress, I should have a PR tomorrow

Issue #28, opened 2018-08-17
Add accessors to the Message class


abompard:
Fedmsg has a lot of accessors available for consumers, let's discuss what we want to keep.

Here's what I found in `fedmsg-meta`, and my comments:
* `title`: replaced by `summary`
* `subtitle`: replaced by `__str__`
* `subjective`: it hasn't really been used, let's drop it
* `long_form` (paragraphs of text about a message): it would be useful for email notifications
* `lexer` (pygments lexer that can be applied to the long_form): not sure we want that, it's only implemented for github and scm messages
* `link`: that would be useful I think (maybe renamed as `url`? I don't find "link" very clear)
* `icon` and `secondary_icon`: those are usually the app icon and the agent's avatar. They are used, so I think we should keep them, but rename them to `app_icon` and `agent_avatar` just to be clearer. (see `fedmsg_meta_fedora_infrastructure.fasshim`).
* `usernames` (a set of FAS usernames associated with the message): this would be necessary for FMN
* `packages` (a set of package names associated with a message): this would be necessary for FMN
* `objects`: unclear, we should drop it and leave it to publishers to provide a precise accessor in their API.
* `emails`: we can leave that to publishers and not make it standard
* `avatars` (dict of avatar URLs associated with a message): unused, drop it

pypingou:
Could it be that the lexer may be used by datagrepper to show the message with a certain high-lighting?

Is ``subjective`` used by the conglomerating code in FMN?

jeremycline:
> * title: replaced by summary
> * subtitle: replaced by __str__
> * long_form (paragraphs of text about a message): it would be useful for email notifications

Hmm. I was thinking that ``__str__`` would basically be the long-form that you'd normally print to a user. Emails would be ``summary`` as the subject, ``__str__`` as the body. IRC would just be summary. Is this enough granularity?

> * lexer (pygments lexer that can be applied to the long_form): not sure we want that, it's only implemented for github and scm messages

I vote drop here. If those messages want lexed output, they can add their own method/property to do so.

> * link: that would be useful I think (maybe renamed as url? I don't find "link" very clear)

Yes, probably useful for users. +1 on the rename to url, and we can provide a sane default in the base Message.

> * icon and secondary_icon: those are usually the app icon and the agent's avatar. They are used, so I think we should keep them, but rename them to app_icon and agent_avatar just to be clearer. (see fedmsg_meta_fedora_infrastructure.fasshim).

[...]

PR #32, opened 2018-08-20
Add accessors on the Message class

abompard:
Fixes: #28


abompard:
I'll just merge this now because I want to push the Black PR which will reformat everything and cause conflicts everywhere :-)
We can always amend that later.

PR #33, opened 2018-08-21
Use Black

abompard:
This PR sets up Black, the Python code formatter, and runs it on the source code. No other change.


abompard:
Black only runs on Python 3.6, and I mistakenly added it to dev-requirements.txt which caused all other tests to fail. Fixing that.

PR #35, opened 2018-08-21
Add support for client x509 authentication in Twisted

jeremycline:
Create the connection with a client certificate, if configured to do so.

@abompard As an aside, what do you think about taking something like this and getting it into upstream pika? It seems like if the SSL context has client certs set, we should automatically be a) setting auth to EXTERNAL, and b) producing the correct twisted connection parameters from the pika parameters.

### DIFF
--- a/fedora_messaging/twisted/service.py
+++ b/fedora_messaging/twisted/service.py
@@ -89,7 +89,7 @@ class FedoraMessagingService(service.MultiService):
                 host=self._parameters.host,
                 port=self._parameters.port,
                 factory=factory,
-                contextFactory=ssl.ClientContextFactory(),
+                contextFactory=_ssl_context_factory(self._parameters),
             )
         else:
             serv = TCPClient(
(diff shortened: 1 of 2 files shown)


fedora_messaging/twisted/service.py
```diff
@@ -103,3 +103,36 @@ def connect(self):
...
+            client_keypair = fd.read()
+        with open(cert) as fd:
+            client_keypair += fd.read()
+        client_cert = ssl.PrivateCertificate.loadPEM(client_keypair)
+
+    context_factory = ssl.optionsForClientTLS(
+        parameters.host,
+        trustRoot=ca_cert,
+        clientCertificate=client_cert,
+        extraCertificateOptions={"raiseMinimumTo": ssl.TLSVersion.TLSv1_2},
```
puiterwijk: Nice! 🎉

abompard:
To make it generic I would need to get the certificates from pika's `parameter.ssl_options` value (we're currently getting them from our config). It's easy on older versions of pika because it's just a dict, but when it's an [ssl.SSLContext](https://docs.python.org/3/library/ssl.html#ssl.SSLContext) I can't find a way to extract the certificates and feed them to Twisted's `ssl.optionsForClientTLS()` function.

Am I missing something? I looks like it's intentional that I don't get access to the loaded certificates. Are you thinking of another way to make this code generic enough to get it into pika?

jeremycline:
> Am I missing something? I looks like it's intentional that I don't get access to the loaded certificates. Are you thinking of another way to make this code generic enough to get it into pika?

You're not missing something, I made it no further than thinking "it'd be nice if this Just Worked" and a cursory investigation of the pyOpenSSL object before I bailed and loaded things from the config. Given that pyOpenSSL is supposed to be replaced with cryptography (I don't know if Twisted has started on this or not), I don't think it's worth seeing if it's even possible to extract enough information from the object to construct the Twisted object. It'd be nice to revisit at some point, but this isn't a huge amount of boilerplate.

Maybe a nice middle ground for the present is to document what is necessary in pika.

abompard:
Yeah it would be nice to have it documented there. They have a usage example in the `examples/twisted_service.py` file but it's very basic.
In the meantime, this is good to go.

PR #38, opened 2018-08-22
Add mergify config to allow automatic merge when a PR has been approved.

cverna:
Signed-off-by: Clement Verna <cverna@tutanota.com>

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,15 @@
+rules:
+  default:
+    protection:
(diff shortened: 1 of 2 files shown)


abompard:
With explanations from cverna, I now find this actually useful :-)
> <cverna> the main advantage is that it rerun the CI after a merge, so when a PR is merged it was tested against the latest master
> https://julien.danjou.info/how-i-stopped-merging-broken-code/

PR #43, opened 2018-08-23
new cli publish command

sebwoj:

In test_good_conf, is it the good idea to check:
mock_publish.assert_called_with(
            api.Message(topic=None, body={None:None}), None
        )

instead of creating anther patch ->
@mock.patch("fedora_messaging.cli.api.Message")
and as return_value set mock_message, and han check:
mock_publish.assert_called_with(
            mock_mnessage, None
        )


sebwoj:
How to rerun Travis build?

jeremycline:
A quick way is to amend your latest commit without changing anything - ``git commit --amend --no-edit`` - and force pushing to your branch: ``git push -f``

fedora_messaging/cli.py
```diff
@@ -150,3 +167,33 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
                 str(e.reason),
             )
             sys.exit(e.exit_code)
+
+
+@cli.command()
+@click.option("--publish-exchange", help=_publish_exchange_help)
+@click.option("--body-key", help=_body_key_help)
+@click.option("--body-value", help=_body_value_help)
```
jeremycline: I wonder if it would be nicer to accept the path to a file that contains JSON. If we did that, we'd need to come up with a format for a JSON-serialized message (message topic, headers, body, etc). That seems generally useful anyway for users who want to create message test fixtures.

What do you think, @abompard?

abompard: Yeah I agree, that'd be useful anyway.

sebwoj:
I'm not sure about information provided to user in case of 'loads' failure.
@jeremycline could you take a look on them?
Command 'record' is on the way.

fedora_messaging/cli.py
```diff
@@ -155,3 +164,120 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+    """
+    Send messages from provided file to an AMQP queue.
+    """
+    msgs_json_str = file.read()
+
+    try:
+        messages = loads(msgs_json_str)
+    except exceptions.ValidationError as e:
+        _log.error("Unable to validate message: %s", str(e))
+        sys.exit(1)
```
jeremycline: With each of these failure cases exiting as 1, it'll be difficult for scripts to handle specific error scenarios. If a file is invalid, no amount of retrying will help, but if a connection can't be established, waiting a bit might help. I recommend adding different exit codes for the different cases and documenting those codes.

Here I think you can just raise https://click.palletsprojects.com/en/7.x/api/#click.BadArgumentUsage or similar with the message you're logging and it'll do the right thing.

fedora_messaging/cli.py
```diff
@@ -155,3 +164,120 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+        except exceptions.PublishReturned as e:
+            _log.error("Unable to publish message: %s", str(e.reason))
+            sys.exit(1)
+        except exceptions.ConnectionException as e:
+            _log.error("Unable to connect to the queue: %s", str(e.reason))
+            sys.exit(1)
+
+
+@cli.command()
+@click.option("--file", help=_record_file_help, required=True, type=click.File("w"))
```
jeremycline: Again, this might be better as an argument, especially if options don't allow for ``-`` to mean stdin/stdout. That way, users can just provide ``-`` to get messages printed to the console with this command.

fedora_messaging/cli.py
```diff
@@ -155,3 +164,120 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+@click.option("--limit", help=_limit_help, type=click.IntRange(1))
+@click.option("--app-name", help=_app_name_help)
```
jeremycline: Defaulting this to "recorder" might be nice.

fedora_messaging/cli.py
```diff
@@ -155,3 +164,120 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+@click.option("--routing-key", help=_routing_key_help, multiple=True)
+@click.option("--queue-name", help=_queue_name_help)
+@click.option("--exchange", help=_exchange_help)
+def record(exchange, queue_name, routing_key, app_name, limit, file):
+    """Record messages from an AMQP queue to provided file."""
+
+    class Recorder:
```
jeremycline:
```suggestion
    class Recorder:
        """
        A simple callback that records messages.
        
        Attributes:
            counter (int): The number of messages this callback has recorded.
            messages (list): The list of messages received.
        """
```

fedora_messaging/cli.py
```diff
@@ -155,3 +164,120 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+
+            def _signal_handler(signum, frame):
+                """
+                Signal handler that gracefully terminates the recorder
+
+                Args:
+                    signum (int): The signal this process received.
+                    frame (frame): The current stack frame (unused).
+                """
+                if signum in (signal.SIGTERM, signal.SIGINT):
```
jeremycline: :+1:

fedora_messaging/cli.py
```diff
@@ -155,3 +164,120 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+            self.counter += 1
+            if limit and limit == self.counter:
+                self.save_messages_to_file()
+                raise exceptions.HaltConsumer(exit_code=0, requeue=False)
+
+        def save_messages_to_file(self):
+            try:
+                json_str = dumps(self.messages)
+            except TypeError as e:
+                _log.error("Unable to save messages to file: ", str(e))
```
jeremycline: This should result in a non-zero exit code, so I think you can just raise HaltConsumer here with that code.

I also think you'll probably want to write the error with https://click.palletsprojects.com/en/7.x/api/#click.echo

fedora_messaging/cli.py
```diff
@@ -155,3 +164,120 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+                """
+                if signum in (signal.SIGTERM, signal.SIGINT):
+                    self.save_messages_to_file()
+                    raise exceptions.HaltConsumer(exit_code=0, requeue=False)
+
+            signal.signal(signal.SIGTERM, _signal_handler)
+            signal.signal(signal.SIGINT, _signal_handler)
+
+        def __call__(self, message):
+            self.messages.append(message)
```
jeremycline: If you wanted to be super fancy here, you could set up a click progress bar (https://click.palletsprojects.com/en/7.x/api/#click.progressbar) so users can see how things are progressing. That could also be a follow-up feature, I think.

fedora_messaging/cli.py
```diff
@@ -155,3 +164,120 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+    if not bindings:
+        raise click.ClickException(
+            "No bindings are defined in the configuration file"
+            " and none were provided as arguments!"
+        )
+
+    if app_name:
+        config.conf["client_properties"]["app"] = app_name
+
+    _log.info("Recording to file %s started.", file)
```
jeremycline: As mentioned above, you probably want to use click.echo here (or the progress bar)

fedora_messaging/cli.py
```diff
@@ -155,3 +164,120 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+        sys.exit(1)
+    except KeyError as e:
+        _log.error("Unable to create message. Missing attribute %s", str(e))
+        sys.exit(1)
+    except ValueError as e:
+        _log.error("Unable to load serialized message: %s", str(e))
+        sys.exit(1)
+
+    for msg in messages:
+        _log.info("Sending message with topic %s", msg.topic)
```
jeremycline: Rather than logging I think you want to use https://click.palletsprojects.com/en/7.x/api/#click.echo

fedora_messaging/cli.py
```diff
@@ -155,3 +164,120 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+        try:
+            api.publish(msg, publish_exchange)
+        except exceptions.PublishReturned as e:
+            _log.error("Unable to publish message: %s", str(e.reason))
```
jeremycline: Rather than exiting 1 here, I think it'd be good to use a different value than the default "something exploded" error status so scripts can potentially handle error cases differently. It's not a great mapping, but maybe https://docs.python.org/3/library/errno.html#errno.EREMOTEIO?

fedora_messaging/cli.py
```diff
@@ -155,3 +164,120 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+            sys.exit(1)
+        except exceptions.ConnectionException as e:
+            _log.error("Unable to connect to the queue: %s", str(e.reason))
+            sys.exit(1)
```
jeremycline: Same deal here, maybe use https://docs.python.org/3/library/errno.html#errno.ECONNREFUSED?

sebwoj:
Thanks for all tips!
Check out my new version of cli.

fedora_messaging/tests/unit/test_cli.py
```diff
@@ -369,3 +374,135 @@ def test_consume_halt_without_exitcode(self, mock_consume, mock_importlib):
...
+            [
+                "--conf=" + GOOD_CONF,
+                "publish",
+                "--exchange=" + cli_options["exchange"],
+                cli_options["file"],
+            ],
+        )
+        self.assertIn("Sending message with topic test_topic", result.output)
+        mock_publish.assert_called_once()
+        self.assertEqual(mock_publish.call_args_list[0][0][0].topic, "test_topic")
```
jeremycline: It'd be clearer here to assign ``mock_publish.call_args_list[0][0][0]`` to a variable like ``sent_msg`` or something. You should also be able to create an ``expected_msg`` object with the expected body, topic, etc. and just do ``self.assertEqual(expected_msg, sent_msg)`` since the message class implements the equality check.

fedora_messaging/cli.py
```diff
@@ -155,3 +160,38 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+    with click.progressbar(messages, label="Sending messages") as bar:
+        for msg in bar:
+            click.echo("Sending message with topic {}".format(msg.topic))
+            try:
+                api.publish(msg, exchange)
+            except exceptions.PublishReturned as e:
+                click.echo("Unable to publish message: {}".format(str(e.reason)))
+                sys.exit(errno.EREMOTEIO)
+            except exceptions.ConnectionException as e:
+                click.echo("Unable to connect to the queue: {}".format(str(e.reason)))
```
jeremycline: I'd have this be "Unable to connect to the message broker: {}"

fedora_messaging/cli.py
```diff
@@ -195,3 +196,99 @@ def publish(exchange, file):
...
+@click.argument("file", type=click.File("w"))
+@click.option("--limit", help=_limit_help, type=click.IntRange(1))
+@click.option("--app-name", help=_app_name_help, default="recorder")
+@click.option("--routing-key", help=_routing_key_help, multiple=True)
+@click.option("--queue-name", help=_queue_name_help)
+@click.option("--exchange", help=_exchange_help)
+def record(exchange, queue_name, routing_key, app_name, limit, file):
+    """Record messages from an AMQP queue to provided file."""
+
+    class Recorder:
```
jeremycline: So what I'd recommend here is pulling the class out of the function. It's not *wrong*, but I think it's a little unexpected (I usually only define classes and functions in a parent class/function scope it if is both tiny and relevant only to the internal workings of the parent). Defining this class at the module level should be fine.

After that, I think you can get rid of pretty much all the rest of the code in the record command and have it just call the "consume" cli command here. The only tricky bit is the consume call expects the callback to be a path. I think you could either hard-code the path or add a check to the consume code to see if it's already a class/function. The only downside to hard-coding I see is if the module name changes it'll break, but that will also break APIs so I don't think that's a huge concern.

sebwoj: But the question is how to pass limit and file to Record instance(instance of class outside record func). As I remember consumer api is responsible to make instance of class . Mabye classmethod?

jeremycline: Ah, right. Maybe cli.consume() should accept a class/callable then, since you can create an instance of the class with the limit and file and pass it on.

sebwoj:
UT stands for unittests. Since Recorder class was inside recorder func, I have no idea how to test Recorder class(call it in UTs).

fedora_messaging/cli.py
jeremycline: Also a good point. Given those things maybe leaving it as-is is fine. Or, since testing it here is a bit troublesome, perhaps pull it out into a function called ``recorder_factory`` or something which accepts the limit as an argument and returns the class with the limit baked in:

```
def recorder_factory(limit):
    class Recorder:
        def __init__(self):
             self.limit = limit
    return Recorder
```
Perhaps a future enhancement to api.consume would be to allow passing instances of classes. I'd say we shouldn't deal with that right now.

fedora_messaging/tests/unit/test_cli.py
jeremycline: True, although you could set the ``expected_msg._headers["sent-at"]`` to be the same as the fixture message

abompard:
I'd like to revieve that PR if possible, @sebwoj do you still want to work on it?

sebwoj:
Yes. I was busy with Bodhi, but I think that I can handle this PR during the weekend.

fedora_messaging/cli.py
```diff
@@ -168,3 +174,176 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+                is impossible.
+        """
+        # Reset the signal handler
+        for signum in (signal.SIGTERM, signal.SIGINT):
+            signal.signal(signum, signal.SIG_DFL)
+        try:
+            json_str = dumps(self.messages)
+        except TypeError as e:
+            click.echo("Unable to save messages to file: {}".format(str(e)))
+            raise exceptions.HaltConsumer(exit_code=1, requeue=False)
```
abompard: If we can't write the message to the file, maybe we want the message to be requeued here?

fedora_messaging/cli.py
```diff
@@ -168,3 +174,176 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+        else:
+            self._file.write(json_str)
```
abompard: In the functions above, the `HaltConsumer` exception is always raised after calling the `save_messages_to_file()` method (which makes sense since we're resetting signal handlers, we must exit afterwards). Thus you could move the 2 `raise exceptions.HaltConsumer(exit_code=0, requeue=False)` calls at the end of this method. You'll avoid repeating yourself and make sure you don't forget to raise in case you later add one more call to the method.

fedora_messaging/tests/unit/test_cli.py
```diff
@@ -142,7 +148,7 @@ def test_no_cli_bindings(self, mock_consume):
 
     @mock.patch("fedora_messaging.cli.api.consume")
     def test_queue_and_routing_key(self, mock_consume):
-        """Assert providing improper bindings is reported."""
+        """Asser  providing improper bindings is reported."""
```
abompard: Probably a typo here

fedora_messaging/cli.py
jeremycline: If dumping raises a TypeError I believe it's not valid JSON so dropping it seems appropriate to me. Although it seems like this code would never be executed in practice since the consume loop would drop the message when json-schema failed on it, so it would never be passed to the callback.

abompard: Ah right, then the error message above is misleading.

fedora_messaging/cli.py
sebwoj: Good point. I will update docblocks as well.

fedora_messaging/tests/unit/test_cli.py
sebwoj: I noticed it while working on UT for record subcommand. Actually this whole docstr description is wrong, but I retyped it in case of "recorder" test. Since this change will be not related to this PR, I will prepare another PR to keep commits [atomic](https://bodhi.fedoraproject.org/docs/developer/index.html#contribution-guidelines).

abompard:
Hey Seb, I'd like to get that PR merged soonish, could you rebase it to the current HEAD please? I'll review it just afterwards. Many thanks!

jeremycline:
Okay, I've rebased this PR. There's a couple minor issues I've encountered while testing, and one design thought.

The problem is that the ``fedora_messaging.api.consume`` API is behaving differently than the twisted_consume API with regards to exchange declaration. It's trying to declare amq.topic and that's not allowed for the Fedora public broker, so if you try to record from the public broker it doesn't work. Not sure what the exact difference is, but we should fix that.

The second thing is the format publish (and record) uses. I wonder if it would be better to not use a JSON list, but rather a format where it's a list of JSON objects separated by newlines. That way you can do something like:
```
long_running_command_that_wants_to_publish | fedora-messaging publish - 
```
so the long-running command can occasionally write to stdout, one line at a time, and publish will pick it up and keep on doing "stdout.readline()" until it gets an end-of-file. With a JSON list, it won't publish anything until it gets an EOF and then it exits.

If ``fedora-messaging record`` wrote out each message one line at a time, I think you could do ``fedora-messaging record - | while read -r line || -n $line; do echo "message received"; done`` to implement #155 with no additional code.

pypingou:
@sebwoj Did you see the latest comments?

abompard:
I totally +1 the idea of having one message per line, and if possible to have the loads/dumps top-level functions accept an object and not necessarily a list. I've needed that to setup fedora-messaging in Bodhi CI.

sebwoj:
Please take a look at #172.

jeremycline:
I'm in favor of keeping the work in a single PR. FYI, I've got a lot of work for this done in a local branch, but I was waiting to see what people thought of the design before pushing it.

As for backwards compatibility, I don't have a problem breaking it and releasing 2.0. I don't think anyone was using the dumps/loads APIs yet (it's not documented in the API) so it should be a non-event for everyone involved.

jeremycline:
Okay, this has been rebased on top of the crochet work so it should be ready for review after that's merged.

docs/fedora-messaging.rst
```diff
@@ -30,17 +30,28 @@ Options
...
+in detail in their own sections below.
 
 ``fedora-messaging consume [OPTIONS]``
 
     Starts a consumer process with a user-provided callback function to execute
     when a message arrives.
 
+``fedora-messaging publish [OPTIONS] FILE``
+
+    Loads serialized messages from a file and publishes them to the specified
```
dustymabe: could we get an example of what a `serialized message` looks like?

dustymabe:
Currently if I call publish on a file that has newlines in it (i.e. pretty formatted json) it barfs. If I get rid of the newlines it works. 

```
[vagrant@vanilla-f30 fedora-coreos-koji-tagger]$ cat test
{
    "body": {
        "encouragement": "You're doing great!"
    },
    "headers": {
        "fedora_messaging_schema": "base.message",
        "fedora_messaging_severity": 20,
        "sent-at": "null"
    },
    "id": "null",
    "queue": "new",
    "topic": "nice.message"
}
[vagrant@vanilla-f30 fedora-coreos-koji-tagger]$ cat test | fedora-messaging publish -

Error: Unable to validate message: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
[vagrant@vanilla-f30 fedora-coreos-koji-tagger]$ 
[vagrant@vanilla-f30 fedora-coreos-koji-tagger]$ cat test | tr '\n' ' ' | fedora-messaging publish - 
Publishing message with topic nice.message
```

jeremycline:
I added some docs to the man page about the message format expectations.

docs/fedora-messaging.rst
jeremycline: I added a description down in the detailed section below. Please let me know if there's anything that isn't clear in the description.

docs/fedora-messaging.rst
```diff
@@ -91,8 +102,95 @@ configuration file and no options on the command line.
...
+
+    Setting this option is equivalent to setting the ``exchange`` setting
+    in *all* ``bindings`` entries in the configuration file.
+
+
+publish
+-------
+
+The publish command expects the message or messages provided in ``FILE`` to be
+JSON objects with each message separated by a newline character. The JSON
```
dustymabe: oh wow.. the newline was required as a delimeter.. did not know that.

docs/fedora-messaging.rst
```diff
@@ -91,8 +102,95 @@ configuration file and no options on the command line.
...
+        },
+        "id": {"type": "string", "description": "The message's UUID."},
+        "body": {"type": "object", "description": "The message body."},
+        "queue": {
+            "type": "string",
+            "description": "The queue the message arrived on, if any.",
+        },
+    },
+    "required": ["topic", "headers", "id", "body", "queue"],
+  }
```
dustymabe: this is great! I think it would be even better if we had a real example below (i.e. we pass in a real world example message into the `fedora-messaging publish -`

dustymabe: I went back and look at [my comment from yesterday](https://github.com/fedora-infra/fedora-messaging/pull/43#issuecomment-506059053).. we could use that example if we wanted (or tweak it however you'd like)

jeremycline:
On Thu, Jun 27, 2019 at 07:38:17PM -0700, Dusty Mabe wrote:
> > @@ -91,8 +102,95 @@ configuration file and no options on the command line.
>  ``--exchange``
>  
>      The name of the exchange to bind the queue to. Can contain ASCII letters,
> -    digits, hyphen, underscore, period, or colon. If one is not specified, the
> -    default is the ``amq.topic`` exchange.
> +    digits, hyphen, underscore, period, or colon.
> +
> +    Setting this option is equivalent to setting the ``exchange`` setting
> +    in *all* ``bindings`` entries in the configuration file.
> +
> +
> +publish
> +-------
> +
> +The publish command expects the message or messages provided in ``FILE`` to be
> +JSON objects with each message separated by a newline character. The JSON
> 
> oh wow.. the newline was required as a delimeter.. did not know that. 
> 

Yeah, this isn't great, but the reason was that I wanted you to be able
to pipe messages in and they'd be published as they were piped.
Requiring everything on one line makes it easy to just loop over
incoming lines and parse them as JSON.

If you've got a better way to do it I'd love to hear it since I'm not
crazy about the current approach either.

dustymabe:
>  If you've got a better way to do it I'd love to hear it since I'm not crazy about the current approach either.

maybe since the messages are json you can just treat one message as the start and close of curly braces `{}`. Start reading a message at `{` and finish at `}`, repeat.

jeremycline:
On Fri, Jun 28, 2019 at 07:14:23AM -0700, Dusty Mabe wrote:
> >  If you've got a better way to do it I'd love to hear it since I'm not crazy about the current approach either.
> 
> maybe since the messages are json you can just treat one message as the start and close of curly braces `{}`. Start reading a message at `{` and finish at `}`, repeat. 
> 

Perhaps this is safe:

```
message = ''
while True:
    message.append(f.readline())
    try:
        json.loads(message)
        break
    except ValueError:
        continue
```

I can't think of a message where this would exit the loop before getting
the complete message, but I haven't thought about it very hard either...

I guess it would fail schema validation if that happened anyway. I'll
experiment a bit.

jeremycline:
I did think of a serious flaw for that approach. A user puts in some malformed JSON, and now you never exit that loop and you don't know if the user messed up or if it's just going to be right with the next line.

The one message per line approach isn't pretty, but at least it lets you know when you're doing it wrong.

abompard:
Yep, that looks good, anything else you want to add / change @jeremycline ?

jeremycline:
I think I'm happy with it (and we can improve it as we go, obviously). I've dismissed my old review which was blocking the merge, rebased, and I'll go ahead and re-approve this myself so the bot merges it when the tests are done.

PR #46, opened 2018-08-26
add unittests for cli consume

sebwoj:

Any hints/tips/corrections related to my work are welcome.

### DIFF
--- a/fedora_messaging/tests/unit/test_cli.py
+++ b/fedora_messaging/tests/unit/test_cli.py
@@ -22,7 +22,7 @@ import unittest
 from click.testing import CliRunner
 import mock
 
-from fedora_messaging import cli
+from fedora_messaging import cli, exceptions
 
 FIXTURES_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "../fixtures/"))
 GOOD_CONF = os.path.join(FIXTURES_DIR, "good_conf.toml")
@@ -47,11 +47,13 @@ class BaseCliTests(unittest.TestCase):
 class ConsumeCliTests(unittest.TestCase):
     """Unit tests for the 'consume' command of the CLI."""
 
+    def setUp(self):
+        self.runner = CliRunner()
+
     @mock.patch("fedora_messaging.cli.api.consume")
     def test_good_conf(self, mock_consume):
         """Assert providing a configuration file via the CLI works."""
-        runner = CliRunner()
-        result = runner.invoke(cli.cli, ["--conf=" + GOOD_CONF, "consume"])
+        result = self.runner.invoke(cli.cli, ["--conf=" + GOOD_CONF, "consume"])
         mock_consume.assert_called_with(
             echo, [{"exchange": "e", "queue_name": "q", "routing_key": "#"}]
         )
@@ -60,8 +62,7 @@ class ConsumeCliTests(unittest.TestCase):
     @mock.patch("fedora_messaging.cli.api.consume")
     def test_conf_env_support(self, mock_consume):
         """Assert FEDORA_MESSAGING_CONF environment variable is supported."""
-        runner = CliRunner()
-        result = runner.invoke(
+        result = self.runner.invoke(
             cli.cli, ["consume"], env={"FEDORA_MESSAGING_CONF": GOOD_CONF}
         )
         mock_consume.assert_called_with(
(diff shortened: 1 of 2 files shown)


fedora_messaging/tests/unit/test_cli.py
```diff
@@ -76,17 +77,219 @@ def test_bad_conf(self, mock_consume):
...
             "Error: Invalid value: thispathdoesnotexist is not a file", result.output
         )
+        """
+        @click.option("--app-name", help=_app_name_help)
+        @click.option("--callback", help=_callback_help)
+        @click.option("--routing-key", help=_routing_key_help)
+        @click.option("--queue-name", help=_queue_name_help)
+        @click.option("--exchange", help=_exchange_help)
+        @click.option("--amqp-url", help=_amqp_url_help)
+        """
```
jeremycline: I don't think this does anything. Did you mean to put this here?

fedora_messaging/tests/unit/test_cli.py
```diff
@@ -76,17 +77,219 @@ def test_bad_conf(self, mock_consume):
...
+        error_message = "not enough values to unpack (expected 2, got 1)"
+        mock_mod_with_callable = mock.Mock(spec=["callable"])
+        mock_importlib.import_module.return_value = mock_mod_with_callable
+        result = self.runner.invoke(
+            cli.cli, ["consume", "--callback=" + cli_options["callback"]]
+        )
+        mock_importlib.import_module.not_called()
+        mock_consume.assert_not_called()
+        self.assertIn(
+            "Unable to parse the callback path ({}); the "
```
jeremycline: Unfortunately it looks like the unpacking error message changes from 3.4 to 3.5. One way to fix this would be to change the assertions in some way to deal with it, but maybe displaying the nitty-gritty error to the user isn't useful.

Perhaps it'd be better to change this error message to just say ``'Unable to parse the callback path "{provided_callback_path}"; the expected format is "my_package.module:callable_object"' ``. Having the provided path and the expected path format right next to each other should be enough for someone to figure out what's wrong.

fedora_messaging/tests/unit/test_cli.py
sebwoj: I forgot to remove my "helpfull lines" :)

fedora_messaging/tests/unit/test_cli.py
sebwoj: I notice this after "push", in Travis output. I forgot to test in pyrhon 2.7.
There is a second way to test it(for the future testing and for education purposes):
line -> callback_path = callback or config.conf["callback"]
can return Mock, which can in line
module, cls = callback_path.strip().split(":")
throw ValueError with custom err message.
But I think that your solution will improve portability.
What do you think?

jeremycline: Let's change the error message :+1:

sebwoj:
Weird, when I run:
python3 -m unittest test_cli.ConsumeCliTests
test -> test_consume_halt_with_exitcode fails.
But when I run:
python3 -m unittest test_cli.ConsumeCliTests.test_consume_halt_with_exitcode
test -> test_consume_halt_with_exitcode is OK.
Am I missing something? Maybe it is flushing kind of problem.

jeremycline:
Ah, I think it's that when it exits with a special error code it's using the logging module, so the output won't necessarily go to click's output? I believe the reason I used logging over raising a click exception (which would be desirable) was the inability to give it an arbitrary exit code. Thus, it logs and calls sys.exit itself.

I can look at it more thoroughly tomorrow.

sebwoj:
I agree that "printing messages" via logging module is more natural than click exceptions.
But weird think is that when I used _log.error in "cli publish", both tests(single test and whole test suite inPublishCliTests) worked (all err messages are printed in result.output via _log.error). In case of ConsumeCliTests, single test(test_consume_halt_with_exitcode) passed (result.output contains proper desired err message), but whole test suite fails because desired message is missing in this particular test. 
So, do do you want me to remove assertion on printed message from test:  test_consume_halt_with_exitcode?

Btw, just for curiosity, how did you forced Travis to make build shows "all tests pass" without changing the code?

jeremycline:
> Btw, just for curiosity, how did you forced Travis to make build shows "all tests pass" without changing the code?

By default GitHub gives write access to maintainers of a project if the branch is a PR, so I amended your commit slightly to check the log output.

PR #48, opened 2018-08-27
Add a severity to messages

jeremycline:
Each message can now have a severity associated with it, both at a class
level and on a message-by-message basis.

### DIFF
--- a/fedora_messaging/message.py
+++ b/fedora_messaging/message.py
@@ -34,6 +34,25 @@ import pytz
 from .exceptions import ValidationError
 
 
+#: Indicates the message is for debugging or is otherwise very low priority. Users
+#: will not be notified unless they've explicitly requested DEBUG level messages.
+DEBUG = 10
+
+#: Indicates the message is informational. End users will not receive notifications
+#: for these messages by default. For example, automated tests passed for their
+#: package.
+INFO = 20
+
+#: Indicates a problem or an otherwise important problem. Users are notified of
+#: these messages when they pertain to packages they are associated with by default.
+#: For example, one or more automated tests failed against their package.
+WARNING = 30
+
+#: Indicates a critically important message that users should act upon as soon as
+#: possible. For example, their package no longer builds.
+ERROR = 40
+
+
 _log = logging.getLogger(__name__)
 
 # Maps regular expressions to message classes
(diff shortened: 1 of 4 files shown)


jeremycline:
> While I believe that the specific values of severities might be hard to determine, this code looks reasonable to me.

Do you mean for a particular message it's difficult to say what the severity is? I agree and I expect we'll have plenty of spirited discussions stemming from someone getting notifications they're not interested it. It's not perfect, but hopefully that will be a guiding force for the "right" severity level.

@abompard, I think I'm going to go ahead and merge this for a beta release (which includes all the fixed-up auth stuff), but when you get back from PTO, please review this and we can adjust it before a stable release is tagged.

fedora_messaging/message.py
```diff
@@ -288,18 +340,18 @@ def validate(self):
...
+                'Validating message headers "%r" with schema "%r"',
+                self._headers,
+                schema,
+            )
+            jsonschema.validate(self._headers, schema)
+        for schema in (self.body_schema, Message.body_schema):
+            _log.debug(
+                'Validating message body "%r" with schema "%r"', self._body, schema
+            )
+            jsonschema.validate(self._body, schema)
```
abompard: You're doing this so people don't override the base schemas too much? Are you thinking of a specific use case?

fedora_messaging/tests/unit/test_message.py
```diff
@@ -22,11 +22,44 @@
...
+        self.assertRaises(
+            exceptions.ValidationError, message.get_message, "", msg._properties, b"{}"
+        )
+
+    def test_missing_headers(self):
+        """Assert missing headers results in a default message."""
+        msg = message.Message()
+        msg._headers = None
+        expected_message = message.Message()
+        expected_message.id = msg.id
```
abompard: You don't seem to be using this variable.

fedora_messaging/message.py
jeremycline: I was thinking that for the headers, it'd be good to validate we added the expected schema, timestamp, etc. keys and the easiest way would be to add it to the schema, but then I wanted people to have the ability to expand upon it. With the body I think it's reasonable to demand all messages be an object, so this makes it enforced.

I can't really think of a use-case for users to actually add headers at the moment so maybe that should all be an internal thing, though.

fedora_messaging/tests/unit/test_message.py
jeremycline: Ah, good eye

jeremycline:
> This looks good! Should the tutorial be updated to ask people to set the default severity?

I'll do that

PR #49, opened 2018-08-28
Remove the unused --amqp-url parameter from the CLI

### LINKED ISSUE #45: "--amqp-url" parameter unused
In cli, parameter "--amqp-url" is not used.
User is not able to pass "amqp-url" to  _session.ConsumerSession(except config file)
Also, impossible to pass "amqp-url" via api.publish to _session.PublisherSession(except config file)

jeremycline:
At the moment the only way to specify the AMQP URL is via the
configuration file. Until someone asks for the ability to specify it on
the CLI, just stick with that.

fixes #45

### DIFF
--- a/docs/fedora-messaging.rst
+++ b/docs/fedora-messaging.rst
@@ -73,11 +73,6 @@ consume
     digits, hyphen, underscore, period, or colon. If one is not specified, the
     default is the ``amq.topic`` exchange.
 
-``--amqp-url``
-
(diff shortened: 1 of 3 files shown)


sebwoj:
For code clarity purposes, we can remove entry ->"amqp_url": "a"<- from unittest in each  @mock.patch.dict

jeremycline:
Indeed. Thanks for looking at it!

PR #54, opened 2018-09-02
unittests for api

sebwoj:

### DIFF
--- a/fedora_messaging/tests/unit/test_api.py
+++ b/fedora_messaging/tests/unit/test_api.py
@@ -1,0 +1,225 @@
+# coding: utf-8
+
+# This file is part of fedora_messaging.
+# Copyright (C) 2018 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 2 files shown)


fedora_messaging/tests/unit/test_api.py
```diff
@@ -0,0 +1,165 @@
...
+    @mock.patch("fedora_messaging.api.pre_publish_signal")
+    @mock.patch("fedora_messaging.api.publish_failed_signal")
+    @mock.patch("fedora_messaging.api.publish_signal")
+    @mock.patch("fedora_messaging._session.PublisherSession")
+    def test_publish_to_exchange(
+        self,
+        mock_publisher_session,
+        mock_publish_signal,
+        mock_publish_failed_signal,
+        mock_pre_publish_signal,
```
jeremycline: I think you can not mock out the signals and instead write a signal handler which calls self.fail() for ones that shouldn't be called. My general feeling is mocking should be for things that can't easily work in the test environment (e.g. the broker connection), but the signals should be simple to use as part of your tests.

sebwoj:
Thanks for all tips.
I missed somehow patch start/stop pattern.
Also have read deeply about unit tests an your approach @jeremycline is right.
Please check out my updated version of UT.

jeremycline:
Looks good to me, I added one small adjustment to the API to explicitly check types. @abompard, what do you think?

PR #64, opened 2018-09-07
Check if consumer session callback class is callable

sebwoj:


fedora_messaging/tests/unit/test_session.py
```diff
@@ -274,12 +274,22 @@ def stop_consumer():
...
+                pass
+
             self.consumer.consume(callback)
             self.assertEqual(self.consumer._consumer_callback, callback)
             connection.ioloop.start.assert_called_once()
-            # Callback is a class
+            # Callback is a callable class
             self.consumer.consume(mock.Mock)
             self.assertTrue(isinstance(self.consumer._consumer_callback, mock.Mock))
+            # Callback is a not callable class
```
jeremycline: Can you break this out into its own test method? That way if this breaks the ``test_consume_uncallable_class`` test case rather than the ``test_consume`` (which I'd expect to be the successful case).

PR #66, opened 2018-09-08
Unittests for session module

sebwoj:


fedora_messaging/tests/unit/test_session.py
```diff
@@ -110,6 +122,28 @@ def test_publish(self):
...
+    def test_publish_without_delivery_confirmation(self):
+        """Check that the publication without delivery confirmation works properly."""
+        self.publisher._confirms = False
+        self.publisher._connection = None
+        self.publisher._channel = None
+        connection_class_mock = mock.Mock()
+        connection_mock = mock.Mock()
+        channel_mock = mock.Mock()
+        connection_class_mock.return_value = connection_mock
+        connection_mock.channel.return_value = channel_mock
```
jeremycline: For future reference, I think you can replace all this with a single connection class mock object, then you can get to the connection and channel by doing ``conn_class_mock.return_value`` and ``conn_class_mock.return_value.channel.return_value`` respectively. Mock objects return mock objects for any attribute or function called. You can also give Mock objects a spec argument so they're explode if an attribute that isn't there on the spec object is accessed. 

I'm happy to accept this as-is since it's testing we didn't have before, but just an FYI :smiley_cat:

sebwoj: Thank you for your tip :)
I just wanted new tests to be more consistent with others.

PR #67, opened 2018-09-10
Use towncrier to generate the changelog

abompard:
Towncrier lets us generate release notes in a much more convenient way than what we currently do. Each PR just has to create a file with the line that should end up in the release notes, and the ``changelog.rst`` file can be automatically generated before each release.

I have added documentation in the ``contributing.rst`` file for more usage info.

@jeremycline I would also like you to verify that the releasing process I have documented in ``contributing.rst`` is correct, please.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -55,6 +55,31 @@ Your pull request should contain tests for your new feature or bug fix. If
 you're not certain how to write tests, we will be happy to help you.
 
 
+Release notes
+-------------
+To add entries to the release notes, create a file in the ``news`` directory
+with the ``source.type`` name format, where ``type`` is one of:
+
+* ``feature``: for new features
+* ``bug``: for bug fixes
+* ``api``: for API changes
+* ``dev``: for development-related changes
+* ``author``: for contributor names
+* ``other``: for other changes
+
+And where the ``source`` part of the filename is:
+
+* ``42`` when the change is described in issue ``42``
+* ``PR42`` when the change has been implemented in pull request ``42``, and
+  there is no associated issue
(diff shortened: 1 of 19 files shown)


docs/contributing.rst
```diff
@@ -101,3 +126,23 @@ address, indicating that you agree to the `Developer Certificate of Origin
...
 Use ``git commit -s`` to add the Signed-off-by tag.
+
+
+Releasing
+---------
+
+When cutting a new release, follow these steps:
+
+* update the version in ``fedora_messaging/__init__.py``
+* update the version in ``docs/conf.py``
```
jeremycline: I have some fanciness to pull the version in the docs from __init__.py, I'll push a commit to this PR

news/get-authors.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
```
jeremycline: Do we need to run this before towncrier to generate the authors for a release? If so it'd be good to note that in the release documentation. It'd be cool to auto-generate the contributors to issues and PRs by looking at the news items and querying the PRs/issues for users as well, since bug reports are valuable contributions.

docs/contributing.rst
abompard: That that looks cool. So this line isn't necessary anymore.

news/get-authors.py
abompard: My original intention was that new `.author` files would be added as part of the PR (just like the news file). But I left this script in case we forgot people and wanted to check before the release.
I agree it'd be cool to get contributors from the GitHub API too, I'll see if I can add it in the future.

jeremycline: Cool, sounds good to me :smile:

PR #72, opened 2018-09-14
Add "consume()" and "cancel()" to the Twisted API

jeremycline:
Allow users to add and remove consumers to a Twisted connection. As a side-effect, each consumer can have its own callback.

I'm planning on a follow-up PR that gives each consumer its own channel to work with so one consumer stopping/messing up doesn't cancel all active consumers.


jeremycline:
Hey @abompard, I've done a bit more adding to the API here, but I'd like your thoughts before I start writing tests (and fixing the rest of the existing ones I broke). It's not broken out into proper commits or anything like that, so I'd recommend just looking at the function signatures and letting me know what you think. I'll tidy it up tomorrow, but I expect you'll wake up before I do :smile:.

The general idea is to stuff the config a user wants to persist across connections into the factory, and push everything else into the protocol class. That way a user sets up the factory with some consumers and can rest easy knowing they'll be automatically reconfigured when connection failures occur.

fedora_messaging/twisted/protocol.py
```diff
@@ -288,13 +476,15 @@ def resumeProducing(self):
...
-        for queue_name in self._queues:
-            queue_object, _consumer_tag = yield self._channel.basic_consume(
-                queue=queue_name
+        for consumer in self._consumers.values():
+            queue_object, _ = yield consumer.channel.basic_consume(
+                queue=consumer.queue, consumer_tag=consumer.tag
+            )
+            deferred = self._read(queue_object, consumer)
+            deferred.addErrback(
+                lambda f: _log.failure, "_read failed on consumer {c}", c=consumer
```
abompard: These 7 lines are duplicated in `consume()` but I'm not sure it's worthwhile to factor them out ;-)

fedora_messaging/twisted/protocol.py
```diff
@@ -66,102 +80,95 @@ def __init__(self, parameters, confirms=True):
...
-        self._channel = yield self.channel()
-        log.msg("AMQP channel created", system=self.name, logLevel=logging.DEBUG)
+        self._channel = yield self._allocate_channel()
         yield self._channel.basic_qos(
             prefetch_count=config.conf["qos"]["prefetch_count"],
             prefetch_size=config.conf["qos"]["prefetch_size"],
+            all_channels=True,
         )
         if self._confirms:
             yield self._channel.confirm_delivery()
```
abompard: This has already been done in `_allocate_channel()`.

fedora_messaging/twisted/protocol.py
```diff
@@ -265,16 +254,215 @@ def publish(self, message, exchange):
...
+            yield consumer.channel.basic_cancel(consumer_tag=consumer.tag)
+        except pika.exceptions.AMQPChannelError:
+            # Consumers are tied to channels, so if this channel is dead the
+            # consumer should already be canceled (and we can't get to it anyway)
+            pass
+        except KeyError:
+            return
+
+        try:
+            yield consumer.channel.close()
```
abompard: Until every consumer is in their own channel, this will close the main channel.

fedora_messaging/tests/integration/test_publish_consume.py
```diff
@@ -5,11 +5,21 @@
 import threading
 import time
 import unittest
+import uuid
 import socket
 
+from crochet import setup, wait_for
```
abompard: Why not use pytest_twisted as elsewhere? Did you have issues with it?

jeremycline: I wanted to run the actual reactor and start real connections, I think I just assumed twisted_pytest wasn't doing that, but it does seem to just use the normal reactor. I can re-write the tests to use that, I think.

jeremycline: Oh, maybe I wanted timeouts so tests wouldn't hang forever? I don't remember.

abompard: If you think using crochet is better, let's keep it and convert the other tests.

abompard: In the past I've seen tests pass when they should fail with several of the tools that try to integrate Twisted with unit tests runners. As a result I'm always a bit worried on that subject, we should be very careful that we're using them properly and not causing false negatives.

jeremycline:
Tests should be fixed up so this is ready for re-review

fedora_messaging/twisted/protocol.py
abompard: @jeremycline what do you think of that comment?

jeremycline: Ah, each consumer now has its own channel, allocated in the ``consume()`` call.

abompard: Do they? I'm not seing this, could you point me to the right line?

abompard: Ah, my bad, the Github UI was somehow showing an old version.

PR #78, opened 2018-09-22
Adjust channel and connection on close callbacks to pika 0.12

sebwoj:
From:

https://pika.readthedocs.io/en/stable/modules/connection.html#pika.connection.Connection.add_on_close_callback

Info:
The callback will be passed the connection, the reply_code (int) and the reply_text (str), if sent by the remote server

Can I assume that all of them (connection, the reply_code, the reply_text) remote server can skip, or just reply_text?


sebwoj:
I agree.
I forgot to check development pika branch.
Now I see that api will change.
Thanks for clarification :)

PR #83, opened 2018-09-25
Add unittest for _session.ConsumerSession

sebwoj:
Missing unittests for ConsumerSession signal handling and methods: _on_cancelok, _on_channel_open, _on_connection_open, _on_channel_close and _on_connection_close.


fedora_messaging/tests/unit/test_session.py
```diff
@@ -292,6 +293,30 @@ def test_external_auth(self):
             consumer._parameters.credentials, credentials.ExternalCredentials
         )
 
+    def test_url_with_client_properties(self):
+        """Assert when URL contains client props, values are not taken from config."""
+        test_url = "amqps://username:password@rabbit.example.com/vhost?client_properties={'k':'v'}"
+        tls_conf = {
```
jeremycline: Since this test is all about the client properties in the URL, I think you can drop the tls config and use an "amqp://..." URL, and drop the assertions about the host, port, and SSL options. Having a test for each of those things is nice because when it fails it's very obvious what part of the URL parsing changed or is broken. The fewer assertions in a test case, the better (generally speaking).

fedora_messaging/tests/unit/test_session.py
```diff
@@ -515,6 +540,194 @@ def test_connection_error_with_pika_exception_as_error(self):
...
+
+        mock_log.info.assert_called_once_with(
+            "Halting %r consumer sessions", consumer_tags
+        )
+        self.assertEqual(self.consumer._running, False)
+        mock_connection.close.assert_called_once_with()
+        self.assertEqual(signal.getsignal(signal.SIGINT), signal.SIG_DFL)
+
+    def test_signal_SIGWINCH(self):
+        """Assert ConsumerSession._shutdown is called on signal SIGIN."""
```
jeremycline: I assume this is supposed to say "Assert ConsumerSession._shutdown isn't called on SIGWINCH."

I think this could be replaced with a test that iterates through all the defined signals except SIGINT and SIGTERM and asserts that they are set to the default handler.

PR #84, opened 2018-09-25
Add abbility to serialize and deserialize messages

sebwoj:

### DIFF
--- a/fedora_messaging/message.py
+++ b/fedora_messaging/message.py
@@ -214,6 +214,114 @@ def get_message(routing_key, properties, body):
     return message
 
 
+def dumps(messages):
+    """
+    Serialize messages to a JSON formatted str
+
+    Args:
+        messages (list): The list of messages to serialize. Each message in
+            the messages is subclass of Messge.
+
+    Returns:
+        str: Serialized messages.
+
+    Raises:
+        TypeError: If at least one message is not instance of Message class or subclass.
+    """
+    serialized_messages = []
+    try:
+        for message in messages:
+            message_dict = message.dump()
+            serialized_messages.append(message_dict)
+    except AttributeError:
+        _log.error("Improper object for messages serialization.")
+        raise TypeError("Message have to be instance of Message class or subclass.")
+
+    return json.dumps(serialized_messages)
+
+
+def loads(serialized_messages):
+    """
+    Deserialize messages from a JSON formatted str
+
+    Args:
+        serialized_messages (JSON str):
+
+    Returns:
+        list: Deserialized message objects.
+
+    Raises:
+        ValidationError: If deserialized message validation failed.
+        KeyError: If serialized_messages aren't properly serialized.
+        json.JSONDecodeError: If serialized_messages is not valid JSON
+    """
+    try:
+        messages_dicts = json.loads(serialized_messages)
+    except json.JSONDecodeError:
+        _log.error("Loading serialized messages failed.")
+        raise
+
+    messages = []
+    for message_dict in messages_dicts:
+        try:
+            headers = message_dict["headers"]
+        except KeyError:
+            _log.error("Message saved without headers.")
+            raise
+
+        try:
+            MessageClass = get_class(headers["fedora_messaging_schema"])
+        except KeyError:
+            _log.error("Message (headers=%r) saved without a schema header.", headers)
+            raise
+
+        try:
+            body = message_dict["body"]
+        except KeyError:
+            _log.error("Message saved without body.")
+            raise
+
+        try:
+            id = message_dict["id"]
+        except KeyError:
+            _log.error("Message saved without id.")
+            raise
+
+        try:
+            queue = message_dict["queue"]
+        except KeyError:
+            _log.warning("Message saved without queue.")
+            queue = None
+
+        try:
+            topic = message_dict["topic"]
+        except KeyError:
+            _log.error("Message saved without topic.")
+            raise
+
+        try:
+            severity = headers["fedora_messaging_severity"]
+        except KeyError:
+            _log.error("Message saved without a severity.")
+            raise
+
+        message = MessageClass(
+            body=body, topic=topic, headers=headers, severity=severity
+        )
+        try:
+            message.validate()
+            _log.debug("Successfully validated message %r", message)
+        except jsonschema.exceptions.ValidationError as e:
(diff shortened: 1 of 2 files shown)


fedora_messaging/message.py
```diff
@@ -186,6 +186,112 @@ def get_message(routing_key, properties, body):
...
+        ]
+    except json.JSONDecodeError as e:
+        raise MsgLoadError(e)
+
+    messages = []
+    for msg_json in deserialized_messages:
+        try:
+            headers = msg_json["headers"]
+        except KeyError:
+            _log.error("Message saved without headers.")
```
jeremycline: I'm inclined to say this should just be an exception - whatever got handed in is not in the format we expect.

fedora_messaging/message.py
```diff
@@ -186,6 +186,112 @@ def get_message(routing_key, properties, body):
...
+            headers = {}
+
+        try:
+            MessageClass = get_class(headers["fedora_messaging_schema"])
+        except KeyError:
+            _log.error("Message (headers=%r) saved without a schema header.", headers)
```
jeremycline: Again, just let this be an exception. Whatever we got aren't properly serialized messages.

fedora_messaging/message.py
```diff
@@ -434,3 +540,17 @@ def packages(self):
...
         return []
+
+    def dump(self):
+        """
+        Dump message attributes.
+
+        Returns:
+            dict: A dictionary of messge attributes.
+        """
+        return {
```
jeremycline: It'd probably be nice to save the message's id as well, so test writers who want to use this functionality can rely on stable message ids.

sebwoj:
PR is updated.
Do you think that exceptions generated by json.dumps(ValueError, TypeError and OverflowError) in 
f-m.message.dumps should be mentioned in method docstring?

fedora_messaging/message.py
```diff
@@ -190,6 +190,115 @@ def get_message(routing_key, properties, body):
...
+
+        try:
+            MessageClass = get_class(headers["fedora_messaging_schema"])
+        except KeyError:
+            _log.error("Message (headers=%r) saved without a schema header.", headers)
+            raise
+
+        try:
+            body = message_dict["body"]
+        except KeyError:
```
jeremycline: I don't think we should default this to something, just let the KeyError propagate.

fedora_messaging/message.py
```diff
@@ -190,6 +190,115 @@ def get_message(routing_key, properties, body):
...
+            _log.error("Message saved without body.")
+            body = {}
+
+        try:
+            id = message_dict["id"]
+        except KeyError:
+            _log.error("Message saved without id.")
```
jeremycline: Also here, messages should always have an id so this should just let the KeyError get raised.

fedora_messaging/message.py
```diff
@@ -190,6 +190,115 @@ def get_message(routing_key, properties, body):
...
+            id = None
+
+        try:
+            queue = message_dict["queue"]
+        except KeyError:
+            _log.error("Message saved without queue.")
+            queue = {}
```
jeremycline: This should probably default to None

fedora_messaging/message.py
```diff
@@ -190,6 +190,115 @@ def get_message(routing_key, properties, body):
...
+
+        try:
+            topic = message_dict["topic"]
+        except KeyError:
+            _log.error("Message saved without body.")
```
jeremycline: This should just let the KeyError get raised

fedora_messaging/message.py
```diff
@@ -190,6 +190,115 @@ def get_message(routing_key, properties, body):
...
+            topic = ""
+
+        try:
+            severity = headers["fedora_messaging_severity"]
+        except KeyError:
+            _log.error("Message saved without a severity. Defaulting to INFO.")
```
jeremycline: I think we should let the KeyError raise here as well. The reason the other code defaults it is because right now all the messages are being sent in fedmsg without it, so we fix things up for that.

fedora_messaging/message.py
```diff
@@ -190,6 +190,115 @@ def get_message(routing_key, properties, body):
...
+        )
+        try:
+            message.validate()
+            _log.debug("Successfully validated message %r", message)
+        except jsonschema.exceptions.ValidationError as e:
+            _log.error("Message validation of %r failed: %r", message, e)
+            raise ValidationError(e)
+
+        message.queue = queue
+        if id:
```
jeremycline: If we assume id will always be set, you can drop this if

jeremycline:
@sebwoj, hey, my guess is the equality method on the exceptions isn't implemented. I don't recall what Python does off the top of my head, but doing ``Exception() == Exception()`` results in False so it could just be checking to see if they are the same instance of the object. I added a commit to just check the message. It looks like there are a couple other test failures, though.

fedora_messaging/message.py
```diff
@@ -510,3 +618,18 @@ def flatpaks(self):
             list(str): A list of affected flatpaks names.
         """
         return []
+
+    def dump(self):
+        """
+        Dump message attributes.
+
+        Returns:
+            dict: A dictionary of messge attributes.
```
jeremycline: "messge" -> "message"

fedora_messaging/message.py
```diff
@@ -510,3 +618,18 @@ def flatpaks(self):
             list(str): A list of affected flatpaks names.
         """
         return []
+
+    def dump(self):
```
jeremycline: I wonder if this is going to be an API most people want to use, or if they should only use the "dumps" and "loads" interfaces outside the class. Maybe we should mark this as private (call it ``_dump``) initially so we don't pollute the namespace with stuff most people won't want to use. What do you think?

sebwoj: As far as I know(c++ exp) private methods are not available outside the class(except friends). So, if we are using "Message.dump" in "dumps" function, it means that we have to allow access from outside. I don't have a big experience in python and I know that underscore(even dunder) can't prevent clients from access private methods, so I'm not sure about final solution. I would like to let you decide in this topic.

jeremycline: Ah, yes, Python has no way to force clients to not touch internals of objects, it's more like guidelines. I think we should make it private (to begin with, anyway) since I can't think of a reason people should be using it.

PR #85, opened 2018-09-26
Only get changelog authors from last tag

Zlopez:
Only small update for get-authors.py script.

This will check git log only from last tag.

### DIFF
--- a/news/get-authors.py
+++ b/news/get-authors.py
@@ -2,9 +2,9 @@
 
 from subprocess import check_output
 
-last_tag = check_output(["git", "describe", "--abbrev=0"])
+last_tag = check_output(["git", "describe", "--abbrev=0"], universal_newlines=True)
 authors = {}
-log_range = last_tag.decode("utf-8").strip() + "..HEAD"
+log_range = last_tag.strip() + "..HEAD"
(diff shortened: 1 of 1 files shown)


news/get-authors.py
```diff
@@ -2,8 +2,10 @@
 
 from subprocess import check_output
 
+last_tag = check_output(["git", "describe", "--abbrev=0"])
 authors = {}
-output = check_output(["git", "log", "--format=%ae\t%an"], universal_newlines=True)
+log_range = last_tag.decode("utf-8").strip() + "..HEAD"
```
abompard: You could also use `universal_newlines=True` to convert the bytes to a string. This is what is used by the other `check_output` call.

Zlopez: Thanks for hint. I will do this.

Zlopez: Still need to strip for '\n'.

Zlopez:
Rebased

abompard:
Thanks. Please run `black` on the file, that's the only failing test now :-)

Zlopez:
Didn't knew, that something like `black` exists. Looks really nice.

abompard:
Yeah it's really cool! There are developer instructions in our docs if you want to use it, but's basically all on their website. There are editor integrations too, I use it with vim & VSCode and it works great.

PR #93, opened 2018-10-05
Set headers for filtering

### LINKED ISSUE #27: Add the metadata for FMN in the headers
The new FMN will leverage RabbitMQ's header exchange feature to distribute relevant message in user-specific queues, that would then be distributed by IRC or email.

To use the header exchange, we need the following data in the headers:
* RPM package name that is affected by the action
* FAS username(s) of the user(s) who is/are affected by the action

We could add the username who did the action, but I'm not sure that's really useful. Why would you want to be notified of actions you take, after all? I could be useful for mentoring, though.

The header exchange in RabbitMQ [only does string matching](https://www.rabbitmq.com/tutorials/amqp-concepts.html#exchange-headers), and there can be multiple users affected by a message. A header won't be matched if it's a list, and there can't be multiple headers with the same key.
It does not appear that we'll be able to entirely rely on RabbitMQ's header routing, and avoid a custom daemon. Am I missing something?

In any case, the base `Message` class must have 2 accessors, `package` and `affected_usernames` (names to be debated) that would return `None` and an empty list by default, and that publishers can override.

jeremycline:
Use the accessors to set message headers which cover the users and content
the message relates to.

Fixes #27

### DIFF
--- a/fedora_messaging/message.py
+++ b/fedora_messaging/message.py
@@ -267,6 +267,7 @@ class Message(object):
         now = datetime.datetime.utcnow().replace(microsecond=0, tzinfo=pytz.utc)
         headers["sent-at"] = now.isoformat()
         headers["fedora_messaging_severity"] = self.severity
+        headers.update(self._filter_headers())
         message_id = str(uuid.uuid4())
         return pika.BasicProperties(
             content_type="application/json",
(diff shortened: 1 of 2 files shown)


abompard:
The `black` test failed, could you fix that please? Apart for that it looks good, thanks! :-)

jeremycline:
Ooops, I've fixed it.

PR #102, opened 2018-11-13
Prepare v1.1.0

jeremycline:


abompard:
Lint test fails, I don't know why this wasn't caught earlier.

jeremycline:
> Lint test fails, I don't know why this wasn't caught earlier.

It looks like the cron job has been failing for 21 days which is when the new flake8 got released, but I never saw an email about it. I think we might need to tweak the travis settings a bit.

PR #103, opened 2018-11-14
Explicitly configure Travis notifications

jeremycline:
The default behavior of Travis is to send notifications to the committer
and the author (but only if they have write access) when a build fails.
This no longer works well because mergify is the committer and the
author may not have write access.

Therefore, explicitly configure Travis to notify us via email when a
build fails. Additionally, it will notify the #fedora-apps channel on
freenode when the build first starts to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,19 @@
 language: python
 cache: pip
 
+notifications:
+  email:
+    recipients:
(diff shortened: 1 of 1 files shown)


abompard:
LGTM, just ignore the lint failure and merge it when you want :-)

PR #108, opened 2018-12-13
Add utilities for message schemas

abompard:
At the moment it's mostly functions to generate avatar URLs.

### DIFF
--- a/fedora_messaging/schema_utils.py
+++ b/fedora_messaging/schema_utils.py
@@ -1,0 +1,72 @@
+# Copyright (C) 2018  Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""
+The ``schema_utils`` module contains utilities that may be useful when writing
+the Python API of your message schemas.
+"""
+
+import collections
+from hashlib import sha256, md5
+
+from six.moves.urllib import parse
+
+
+def user_avatar_url(username, size=64, default="retro"):
+    """Get the avatar URL of the provided Fedora username.
+
+    The URL is returned from the Libravatar service.
+
+    Args:
+        username (str): The username to get the avatar of.
+        size (int): Size of the avatar in pixels (it's a square).
+        default (str): Default avatar to return if not found.
+    Returns:
+        str: The URL to the avatar image.
+    """
+    openid = "http://{}.id.fedoraproject.org/".format(username)
+    return libravatar_url(openid=openid, size=size, default=default)
+
+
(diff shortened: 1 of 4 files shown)


fedora_messaging/schema_utils.py
```diff
@@ -0,0 +1,90 @@
...
+    Args:
+        email (str): the user's email
+        openid (str): the user's OpenID
+        size (int): Size of the avatar in pixels (it's a square).
+        default (str): Default avatar to return if not found.
+        dns (bool): Send a DNS SRV query to use Libravatar federation.
+    Returns:
+        str: The URL to the avatar image.
+    """
+    if dns and libravatar is not None:
```
jeremycline: This makes the API a bit confusing in my opinion. ``dns=True`` implies using libravatar, but only if it's installed, in which case ``dns`` is ignored. This means when I call this function with ``dns=True`` I might get federation, or maybe I won't, but I certainly wont know without inspecting reading the implementation and checking the returned string.

Since all it does is wrap the libravatar API call, I would recommend that we simply document that ``libravatar.libravatar_url`` exists and allows for federation in this function's docblock. If a user wants that, they'll import libravatar themselves and call it. Otherwise they can use this.

fedora_messaging/schema_utils.py
```diff
@@ -0,0 +1,90 @@
...
+        # This makes an extra DNS SRV query, which can slow down our
+        # webapps. It is necessary for libravatar federation, though.
+        return libravatar.libravatar_url(
+            email=email, openid=openid, size=size, default=default
+        )
+    params = collections.OrderedDict([("s", size), ("d", default)])
```
jeremycline: urlencode accepts plain dictionaries, why does this have to be an ordered dictionary?

fedora_messaging/schema_utils.py
```diff
@@ -0,0 +1,90 @@
...
+    query = parse.urlencode(params)
+    if email:
+        idhash = md5(email.encode("utf-8")).hexdigest()
+    elif openid:
+        idhash = sha256(openid.encode("utf-8")).hexdigest()
+    else:
+        raise TypeError("You must provide either the email or the openid.")
```
jeremycline: This isn't documented in the docblock. Also, it's a minor thing, but I think ValueError is more appropriate here since ``None`` is a valid type for both email and openid, but they can't both be None at the same time.

fedora_messaging/schema_utils.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    try:
+        return HARDCODED_AVATARS[username].format(size=size)
+    except KeyError:
+        openid = "http://{}.id.fedoraproject.org/".format(username)
+        return libravatar_url(openid=openid, size=size, default=default)
+
+
+def libravatar_url(email=None, openid=None, size=64, default="retro", dns=False):
+    """Get the user's avatar from libravatar.
```
jeremycline:
```suggestion
    """Get the URL to a user's avatar from libravatar.
```

fedora_messaging/schema_utils.py
```diff
@@ -0,0 +1,90 @@
...
+try:
+    import libravatar
+except ImportError:
+    # We can do without this dependency.
+    libravatar = None
+
+
+# https://github.com/fedora-infra/fedmsg_meta_fedora_infrastructure/issues/320
+HARDCODED_AVATARS = {
+    "bodhi": "https://apps.fedoraproject.org/img/icons/bodhi-{size}.png",
```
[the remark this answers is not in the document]
abompard: Nah you're right, I think I just stared at fedmsg code for too long, that's all.

fedora_messaging/schema_utils.py
abompard: I think it's only to make testing easier.

abompard:
Good point, I've added the docs now :-)

PR #117, opened 2019-01-11
 Allow declaring --exchange/--queue-name/--routing-key independently

jeremycline:
The command-line flags on the consume command are intended to make it
quick and easy to run simple consumers without fiddling with a
configuration file.

To this end, allow users to override --exchange, --queue-name, and
--routing-key, rather than requiring none or all. This brings it in line
with the manual page which states defaults for queue-name and exchange
will be used if they are not provided.

This implementation is simplistic in that it sets all bindings in the
configuration to be for the provided values. However, the command line
interface options do not offer a way exactly declare exchanges, queues,
and bindings (durability options, arbitrary arguments, etc) so the
documentation is updated to indicate for complex queue configurations,
the configuration file should be used instead of command-line arguments.

### DIFF
--- a/fedora_messaging/tests/unit/test_cli.py
+++ b/fedora_messaging/tests/unit/test_cli.py
@@ -140,48 +140,30 @@ class ConsumeCliTests(unittest.TestCase):
             echo, bindings=expected_bindings, queues=config.conf["queues"]
         )
 
-    @mock.patch.dict(
-        "fedora_messaging.config.conf", {"bindings": None, "callback": "mod:callable"}
-    )
-    @mock.patch("fedora_messaging.cli.importlib")
     @mock.patch("fedora_messaging.cli.api.consume")
-    def test_wrong_cli_bindings(self, mock_consume, mock_importlib):
+    def test_queue_and_routing_key(self, mock_consume):
         """Assert providing improper bindings is reported."""
-        cli_options = {"queue-name": "qn", "routing-key": "rk"}
-        mock_mod_with_callable = mock.Mock(spec=["callable"])
-        mock_importlib.import_module.return_value = mock_mod_with_callable
-        result = self.runner.invoke(
-            cli.cli,
-            [
-                "consume",
-                "--queue-name=" + cli_options["queue-name"],
-                "--routing-key=" + cli_options["routing-key"],
-            ],
-        )
-        mock_importlib.import_module.called_once_with("mod")
-        mock_consume.assert_not_called()
-        self.assertIn(
-            "You must define all three of exchange, queue_name and"
-            " routing_key, or none of them to use the configuration",
-            result.output,
-        )
-        self.assertEqual(1, result.exit_code)
-
-    @mock.patch("fedora_messaging.cli.api.consume")
-    def test_missing_cli_and_conf_bindings(self, mock_consume):
-        """Assert missing bindings via cli and in conf is reported."""
-        config.conf["bindings"] = None
         config.conf["callback"] = "fedora_messaging.tests.unit.test_cli:echo"
 
-        result = self.runner.invoke(cli.cli, ["consume"])
-
-        mock_consume.assert_not_called()
-        self.assertEqual(1, result.exit_code)
-        self.assertIn(
-            "No bindings are defined in the configuration file"
-            " and none were provided as arguments!",
-            result.output,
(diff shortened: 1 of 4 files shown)


fedora_messaging/tests/unit/test_cli.py
```diff
@@ -174,32 +170,19 @@ def test_wrong_cli_bindings(self, mock_consume, mock_importlib):
...
-            result.output,
-        )
-        self.assertEqual(1, result.exit_code)
-
-    @mock.patch.dict(
-        "fedora_messaging.config.conf", {"bindings": None, "callback": "mod:callable"}
-    )
-    @mock.patch("fedora_messaging.cli.importlib")
-    @mock.patch("fedora_messaging.cli.api.consume")
-    def test_missing_cli_and_conf_bindings(self, mock_consume, mock_importlib):
```
jeremycline: The diff is a bit confusing here, but I got rid of this test because it tests a scenario that can't happen - when the configuration is loaded, the bindings and queues are checked to make sure they're valid (and not None since the default isn't)

fedora_messaging/tests/unit/test_cli.py
```diff
@@ -130,8 +128,8 @@ def test_good_cli_bindings(self, mock_consume, mock_importlib):
             ],
             queues={
                 "qn": {
-                    "durable": True,
-                    "auto_delete": False,
+                    "durable": False,
```
abompard: Do you think it's better to have `durable` set to `False` by default? Why is that?

jeremycline: I was on the fence here. The reason this changed was that the configuration defaults were non-durable and auto-deleted, but the CLI defaults were durable and not auto-deleted. I can see advantages for both, but it would be good if it matched whatever the configuration default is so using the API and the CLI behave the same.

*Technically* this is a backwards-incompatible change so we should probably bump the version to 2.

abompard: I totally agree with the need for coherence. I don't see the advantage of non-durable, how is it useful?

abompard: Oh I see. Wouldn't "durable & auto-delete" fix this issue? The queue will survive a broker restart (which may happen independently from the client restart) but will be deleted when the client quits? Ah, but if you don't set a queue name and the broker restarts while the client is not restarted, it will reconnect and generate a different queue name, no? Or will it re-use the existing one?

I think setting auto-restart to false is useful when you want messages to keep landing in your queue when you restart a client, but that's more appropriate for well-established clients that we setup in infra, so we can set that in the config file.

Ah, I don't know. Keeping the current defaults has the slight advantage of allowing us to keep the major version number. Which reminds me I haven't set all options in the infra config files, and I should if we want to change those defaults one day.

jeremycline: > I think setting auto-restart to false is useful when you want messages to keep landing in your queue when you restart a client, but that's more appropriate for well-established clients that we setup in infra, so we can set that in the config file.

It generates a random UUID each time the process starts. I see the defaults as being things you don't use, probably not even in a development environment. It's mostly for "I just installed it and want to see what happens when I do <x>". 

> Ah, I don't know. Keeping the current defaults has the slight advantage of allowing us to keep the major version number. Which reminds me I haven't set all options in the infra config files, and I should if we want to change those defaults one day.

Haha, yeah, I kept going back and forth on it and I don't have a good answer. I'm not too concerned about bumping the major version number since it's a super tiny change, it's more I don't know what the best default is. I lean towards the "durable=False, auto-delete=True" because it's tidiest, but "durable=False, auto-delete=False" will clean up on broker restarts so maybe that's the nicest balance. "durable=True, auto-delete=False" is what people will want in production, so an argument could be made that that's what the defaults should be so users don't trip up on that.

abompard: And coïncidentally, that's what our current defaults are! :-) Such good planning.

jeremycline: I've thought about it a bit and I think I'd like to defer changing the queue settings to a different PR. Unfortunately, when changing the tests I noticed they have some ordering issues which I'm going to fix in a separate PR as well, then this PR will be ready.

PR #119, opened 2019-01-23
Make the message body a public property

abompard:
The message `body` property should be public, otherwise schema writers will have to write an accessor for each dictionary entry, which does not make much sense.


abompard:
Yeah, we should. Do you want me to? Anything else to include in that release?

Issue #121, opened 2019-01-24
Add a configurable way to prefix the topic


abompard:
For compatibility with fedmsg, it would be useful to be able to set a topic prefix in the configuration. Something like `org.fedoraproject.prod` (or `staging`).

The existing code does not deal with that by itself since the topic is assembled by fedmsg, and the running code does not know much about the environment it's running in, so the configuration seems like the right place.

PR #122, opened 2019-01-29
Add a configurable way to prefix the topic

abompard:
Fixes: #121

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -7,6 +7,10 @@ publish_exchange = "amq.topic"
 
 callback = "fedora_messaging.example:printer"
 
+# The topic_prefix configuration value will add a prefix to the topics of every sent message.
+# This is used for migrating from fedmsg, and should not be used afterwards.
+topic_prefix = ""
+
 # Note the double brackets below.
 # To add another binding, add another [[bindings]] section.
 [[bindings]]
--- a/fedora_messaging/config.py
+++ b/fedora_messaging/config.py
@@ -130,6 +130,11 @@ publish_exchange
 A string that identifies the exchange to publish to. The default is
 ``amq.topic``.
 
+topic_prefix
+------------
+A string that will be prepended to topics on sent messages.
+This is useful to migrate from fedmsg, but should not be used otherwise.
+The default is an empty string.
 
 .. _sub-config:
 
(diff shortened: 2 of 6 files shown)


fedora_messaging/_session.py
```diff
@@ -180,13 +180,18 @@ def _connect_and_publish(self, exchange, message):
             if self._confirms:
                 self._channel.confirm_delivery()
 
+        routing_key = message._encoded_routing_key
+        if config.conf["topic_prefix"]:
```
jeremycline: Did you find adding this in the Message class to be a problem? I was thinking setting it on message creation so inspection of the message object would show what's actually sent, but I've not finished my coffee yet and maybe I'm not taking something into account.

fedora_messaging/_session.py
```diff
@@ -180,13 +180,18 @@ def _connect_and_publish(self, exchange, message):
...
+            routing_key = (
+                config.conf["topic_prefix"].encode("utf-8") + b"." + routing_key
```
jeremycline: Wherever you choose to make the real routing key, it'll probably be a nicer user experience to make sure the topic prefix doesn't include a trailing ``.``. Something like:

 ``".".join((config.conf["topic_prefix"].rstrip("."), routing_key))``

fedora_messaging/_session.py
abompard: I have tried both, I think each could work, but then I'd rename `_encoded_routing_key` to something like `_final_routing_key` or `_wire_routing_key`, since it's not only an encoding. Any thoughts on that?

fedora_messaging/_session.py
abompard: Indeed, that would avoid accidents.

fedora_messaging/_session.py
jeremycline: Is mutating Message.topic instance variable a problem? Then you could just do ``self.topic = ...`` in ``__init__`` and ``_encoded_routing_key`` would have the right value. In other words, the Message init would look like:

```
def __init__(topic=None, ...):
    if topic:
        # Default is "" on base class
        self.topic = topic

    if config.conf["topic_prefix"]:
        self.topic = ".".join((config.conf["topic_prefix"].rstrip("."), self.topic))
```

I know it's a _bit_ unexpected (I set topic to x, why is it y.x!?), but only if it is explicitly configured.

abompard: Yeah, I thought about that and figured it would feel weird to have an instance attribute with the same name as a constructor argument but with a slightly different value. But yeah, maybe that's just me, I'll do that.

PR #124, opened 2019-01-29
Add a consumer systemd service file

abompard:
Add some documentation to create a systemd service file based on the `consume` subcommand.


abompard:
After discussing it with @jeremycline, I've changed this PR to use a dynamic systemd service file.

PR #135, opened 2019-02-27
Fix the mailman_schema example to work with Message.body

jeremycline:
The message ``body`` variable is used by the Message class, but the
example schema is still defining a property using that name.

See https://pagure.io/fedora-commops/fedora-happiness-packets/issue/96


abompard:
Could you run Black and rebase please?

PR #141, opened 2019-03-12
[Docs] Add instructions for installing the library and RabbitMQ in Docker

ShraddhaAg:
This PR adds pointers for installing fedora-messaging and setting up RabbitMQ in dockerised applications
as suggested by @jeremycline [here](https://pagure.io/fedora-commops/fedora-happiness-packets/issue/96#comment-558547) in the tutorial docs.

### DIFF
--- a/docs/tutorial/installation.rst
+++ b/docs/tutorial/installation.rst
@@ -15,9 +15,11 @@ Create a Python virtual environment::
 
 Install the library and its dependencies::
 
+    pip install fedora-messaging
+    # Alternatively, install it directly from the git repository
     git clone https://github.com/fedora-infra/fedora-messaging.git
     cd fedora-messaging
-    python setup.py develop
+    pip install -e .
 
 Make sure it is available and working::
 
@@ -38,6 +40,9 @@ http://localhost:15672/. The username is ``guest`` and the password is
 read the messages in the queues. Keep it open in a browser tab, we'll need it
 later.
 
+If your project uses containers, consult the `RabbitMQ documentation`_ about containers.
+
+.. _RabbitMQ documentation: https://www.rabbitmq.com/download.html#docker
 
 Configuration
 -------------


docs/tutorial/installation.rst
```diff
@@ -23,6 +23,7 @@ Make sure it is available and working::
 
     fedora-messaging --help
 
+If your project uses Docker, include the library in the ``requirements.txt`` file.
```
jeremycline: I think it'd be best if in the "install the library and its dependencies" section above we just recommend installing it with pip: "pip install fedora-messaging" with a note below that for installing it directly from the git repository. Something like:
```
Install the library and its dependencies::

    pip install fedora-messaging
    # Alternatively, install it directly from the git repository
    git clone https://github.com/fedora-infra/fedora-messaging.git
    cd fedora-messaging
    pip install -e .
```

Not everyone uses ``requirments.txt`` files so I think the hint about pip-installing is enough to give the reader the package name and they can put it in their requirements file or setup.py or Pipenv file or whatever.

@abompard, do you recall why we recommended installing it from source control rather than PyPI?

docs/tutorial/installation.rst
```diff
@@ -38,6 +39,10 @@ http://localhost:15672/. The username is ``guest`` and the password is
 read the messages in the queues. Keep it open in a browser tab, we'll need it
 later.
 
+If your project uses Docker, a conatiner needs to be set up for RabbitMQ.
```
jeremycline: Not everyone uses Docker to manage and run their containers, so I'd recommend phrasing this as:
"If your project uses containers, consult the `RabbitMQ documentation`_ on its container."

ShraddhaAg:
@jeremycline I've incorporated the changes as requested. Please find the time to review :)

docs/tutorial/installation.rst
```diff
@@ -38,6 +40,9 @@ http://localhost:15672/. The username is ``guest`` and the password is
 read the messages in the queues. Keep it open in a browser tab, we'll need it
 later.
 
+If your project uses containers, consult the `RabbitMQ documentation`_ on it's conatiner.
```
abompard: There seem to be a typo on the last word. Also, it's probably a "its" ;-) but you may want to rephrase it as `consult the RabbitMQ documentation about containers`.

docs/tutorial/installation.rst
```diff
@@ -38,6 +40,9 @@ http://localhost:15672/. The username is ``guest`` and the password is
...
+
+.. _RabbitMQ documentation: https://www.rabbitmq.com/download.html#docker?
```
abompard: There is no question mark at the end of the URL if you want to get to the proper target on the page.

ShraddhaAg:
Hi @abompard I've corrected the silly mistakes :sweat_smile:

docs/tutorial/installation.rst
abompard: I don't know, maybe we had no published version at that time.

Issue #143, opened 2019-03-13
The topic_prefix is added to incoming messages


abompard:
The `topic_prefix` is added in the `Message` constructor, as a result it is also added when messages are received from the broker. Since `topic_prefix` is only intended for outgoing messages, this should be fixed.

PR #144, opened 2019-03-13
Only prepend the topic_prefix on outgoing messages

abompard:
Fixes: #143

### DIFF
--- a/fedora_messaging/tests/unit/test_message.py
+++ b/fedora_messaging/tests/unit/test_message.py
@@ -448,10 +448,10 @@ class MessageTests(unittest.TestCase):
         )
 
     def test_topic_prefix(self):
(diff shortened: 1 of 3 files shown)


fedora_messaging/tests/unit/test_message.py
```diff
@@ -447,12 +447,6 @@ def test_body_deprecation_getter(self, mock_warning):
             stacklevel=2,
         )
 
-    def test_topic_prefix(self):
```
jeremycline: It seems like this test can stay, it just should check that the msg._encoded_routing_key is "prefix.test.topic".encode("utf-8")

jeremycline:
@abompard did this get lost in the shuffle? I imagine we'll want to make a 1.6 release soonish, I assume you want this in it.

PR #146, opened 2019-03-21
Add better error handling for permission denied cases

jeremycline:
There are a number of cases which weren't being properly handled on the
less-than-happy path for authentication:

1. If the user had no access to the vhost at all, it would endlessly
   reconnect without any indication of a problem.

2. If authentication failed, it would do the same thing as 1.

3. If the user had access to the vhost, but lacked read permissions, it
   would set up the queue, configure the subscriber, and then would fail
   to read from the queue. The channel would be closed and it would
   assume it was a connection problem, but the connection was fine. The
   consumer would just hang.

For 1 and 2, a ConnectionException is now raised by ``twisted_consume``
as the connection is not going to improve with more retries.

For 3, a new exception has been added, PermissionException (of which
BadDeclaration is now a subclass), which is provided to the
consumer.result errback when there's a permission problem. This gives
the API users an opportunity to do something about it.

Case 3 is not properly handled by Pika 1.0.0b1 and b2 and its test is
skipped in that case. The 403 is never passed back to the reader of the
queue. We need to fix this upstream.

### DIFF
--- a/fedora_messaging/tests/integration/test_twisted_api.py
+++ b/fedora_messaging/tests/integration/test_twisted_api.py
@@ -7,11 +7,12 @@ import time
 
 from twisted.internet import reactor, defer, task
 import treq
+import pkg_resources
 import pytest
 import pytest_twisted
 
 from fedora_messaging import api, message, exceptions, config
-from fedora_messaging.twisted.protocol import _add_timeout
+from fedora_messaging.twisted.protocol import _add_timeout, _pika_version
 
 
 HTTP_API = "http://localhost:15672/api/"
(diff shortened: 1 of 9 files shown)


fedora_messaging/tests/integration/test_twisted_api.py
```diff
@@ -495,12 +496,95 @@ def test_twisted_consume_serverside_cancel():
 
 
 @pytest_twisted.inlineCallbacks
-def test_twisted_consume_permission_denied(admin_user):
+def test_no_vhost_permissions(admin_user):
+    """Assert a hint is given if the user doesn't any access to the vhost"""
```
abompard: missing "have" here.

jeremycline: Fixed!

PR #149, opened 2019-03-26
Update the config docs and add info on configuration for public access

jeremycline:
This has a couple small commits which I didn't feel was worth the effort of breaking out into numerous PRs, but all center around making the configuration documentation clearer, including how to configure access to the Fedora broker.


config.toml.example
```diff
@@ -1,34 +1,22 @@
...
-[[bindings]]
-queue = "my_queue"
-exchange = "amq.topic"
-routing_keys = ["#"]
-
 [tls]
-ca_cert = "/etc/pki/tls/certs/ca-bundle.crt"
-keyfile = "/my/client/key.pem"
-certfile = "/my/client/cert.pem"
+ca_cert = "/etc/fedora-messaging/fedora.cacert.pem"
```
abompard: In the other example config files, this file is called `cacert.pem`. Is it intended to have a different name here?

config.toml.example
```diff
@@ -1,34 +1,22 @@
...
+keyfile = "/etc/fedora-messaging/fedora.key.pem"
+certfile = "/etc/fedora-messaging/fedora.cert.pem"
```
abompard: In the other example config files, those files use a dash between `fedora` and the file type. Is it intended to have a different name here?

configs/fedora.toml
```diff
@@ -0,0 +1,78 @@
+# A basic configuration for Fedora's message broker, using the example callback
+# which simply prints messages to standard output.
+#
+# This file is in the TOML format.
+amqp_url = "amqps://fedora:@rabbitmq.fedoraproject.org/%2Fpublic_pubsub"
+passive_declares = false
```
abompard: Shouldn't that be `true`?

configs/fedora.stg.toml
```diff
@@ -0,0 +1,77 @@
+# A basic configuration for Fedora's staging message broker, using the example
+# callback which simply prints messages to standard output.
+#
+# This file is in the TOML format.
+amqp_url = "amqps://fedora.stg:@rabbitmq.stg.fedoraproject.org/%2Fpublic_pubsub"
+callback = "fedora_messaging.example:printer"
+
```
abompard: Should we set `passive_declares` to `true` here?

configs/fedora.toml
jeremycline: No, it needs to be false because in the public_pubsub vhost users  create their own queues. Since the default is false I think this should be dropped here, though. Looked like I dropped it from the stg config, but not here.

config.toml.example
jeremycline: No, I just adjusted the name repeatedly and they got out of sync. Fixed.

configs/fedora.toml
abompard: Right, I mixed it up. Do you think there may be errors trying to declare the exchange?

jeremycline: It works fine in staging. I'm having an issue in production that looks like a proxy problem - the initial TCP handshake isn't completing, but it's the same configuration.

PR #150, opened 2019-03-26
Add __repr__ implementations to the Twisted factory

jeremycline:
This makes it look more useful when it is logged or printed during
debugging. Before, Twisted emitted the following log:

Starting factory
<fedora_messaging.twisted.factory.FedoraMessagingFactoryV2 object at
0x7f9aa4d4d358>

With this change, the log message becomes:

Starting factory FedoraMessagingFactoryV2(parameters=<URLParameters
host=rabbitmq.stg.fedoraproject.org port=5671
virtual_host=/public_pubsub ssl=True>, confirms=True)

which is much more helpful.

### DIFF
--- a/fedora_messaging/twisted/factory.py
+++ b/fedora_messaging/twisted/factory.py
@@ -286,6 +286,12 @@ class FedoraMessagingFactoryV2(protocol.ReconnectingClientFactory):
         self._client = None
         self._consumers = {}
 
+    def __repr__(self):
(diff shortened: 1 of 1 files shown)


jeremycline:
Please don't approve pull requests that don't have test coverage. I expected this to be covered by tests, but it isn't (see the coverage report).

Zlopez:
Sorry, didn't noticed, that the diff coverage is not checked.

PR #153, opened 2019-04-04
Add PyOpenSSL to the dependency list

jeremycline:
Twisted doesn't depend on PyOpenSSL, but its ssl module does use it. Add
it as an explicit dependency.


jeremycline:
I'm going to go ahead and merge this since it's pretty simple.

PR #162, opened 2019-04-17
Add a "--callback-file" option

### LINKED ISSUE #159: Allow users to provide a filesystem path to a Python file for callbacks
Currently, the callback must be in the PYTHONPATH and is specified by "package.module:function" in the configuration. This is convenient when you already have the function in the Python path and the callback is part of a bigger project.

However, for one-off scripts it's more convenient to provide a path to the file (some/python/file.py:function) than it is to make a Python package for the callback alone.

We should add support to the consume CLI to accept a filesystem path as well as a Python path. It'll need to differentiate the two types. If a relative path is given to a file in the current directory (e.g. "callback.py:my_function") we can't tell the difference between the PYTHONPATH variety and the filesystem path, so I think we should always default to first trying the PYTHONPATH, then falling back to interpreting it as a filesystem path, then failing.

jeremycline:
Previously, the callback needed to reside on the Python Path, which is
convenient if you're already making a Python package. Plenty of folks
aren't, though, so allow users to provide a file that we'll boldly exec
and load a callback from.

Fixes #159

Note: I added this as a new option since I was concerned about messing up distinguishing the filesystem path from the Python path cases. This way the user is explicit, at the expense of things not quite Just Working. Feedback welcome on that decision.

### DIFF
--- a/fedora_messaging/cli.py
+++ b/fedora_messaging/cli.py
@@ -51,6 +51,13 @@ _callback_help = (
     "and should point to either a function or a class. Consult the API "
     "documentation for the interface required for these objects."
 )
+_callback_file_help = (
+    "The path of a Python file that contains the callable object to"
+    " execute when the message arrives. This should be in the format "
+    '"my/script.py:object_in_file". Providing this overrides the callback set '
+    'by the "--callback" argument or configuration file. It executes the entire '
+    "Python file to load the callable object."
+)
 _routing_key_help = (
     "The AMQP routing key to use with the queue. This controls what messages are "
     "delivered to the consumer. Can be specified multiple times; any message "
@@ -91,11 +98,12 @@ def cli(conf):
 
 @cli.command()
 @click.option("--app-name", help=_app_name_help)
+@click.option("--callback-file", help=_callback_file_help)
 @click.option("--callback", help=_callback_help)
 @click.option("--routing-key", help=_routing_key_help, multiple=True)
 @click.option("--queue-name", help=_queue_name_help)
 @click.option("--exchange", help=_exchange_help)
-def consume(exchange, queue_name, routing_key, callback, app_name):
+def consume(exchange, queue_name, routing_key, callback, callback_file, app_name):
     """Consume messages from an AMQP queue using a Python callback."""
     # The configuration validates these are not null and contain all required keys
     # when it is loaded.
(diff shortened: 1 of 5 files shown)


fedora_messaging/tests/fixtures/callback.py
```diff
@@ -0,0 +1,2 @@
+def rand(message):
+    return 4
```
abompard: Hahahahaha https://www.xkcd.com/221/ :rofl:

fedora_messaging/cli.py
```diff
@@ -123,6 +131,94 @@ def consume(exchange, queue_name, routing_key, callback, app_name):
...
+        )
+
+    try:
+        file_namespace = {}
+        with open(file_path, "rb") as fd:
+            try:
+                # Using "exec" is generally a Bad Idea (TM), so bandit is upset at
+                # us. In this case, it seems like a Good Idea (TM), but I might be
+                # wrong. Sorry.
+                exec(compile(fd.read(), file_path, "exec"), file_namespace)  # nosec
```
abompard: I wonder if there would be an alternative way of doing this that would be safer, like maybe modifying the module path and loading it as a module, but I think it would pretty much have the same security implications.

abompard:
I'm not sure I understand your concern in the "Note" paragraph you wrote above, could you elaborate a bit?

jeremycline:
So the first thing I tried was just overloading the "--callback" parameter for the command, and trying to correctly guess if the argument was a filesystem path or a python path. It seemed easy to get right at first, but there are cases where something could be a python path or a filesystem path: "some.module:function" could be either a Python path or a file called "some.module".

It's weird to me to have a "--callback" parameter *and* a "--callback-file" parameter which are both used to specify the callback to use. On the other hand, I like the clarity of specifying if I mean a file or a Python path. The note was just to indicate it's not a design I'm crazy about, but I don't have better ideas.

abompard:
Ah, OK I get it now, thanks. We could use the same option and try each method and see which one works first, but I wonder it that's wise, also from a security point of view.
I don't have a better idea either. Let's add a new option, and if we find the perfect way later we can deprecate it and have the code follow to the improved `--callback` codepath.

fedora_messaging/cli.py
jeremycline: At the very least @puiterwijk should look at it and tell me how bad an idea this is. I can't think if any reasonable (or unreasonable) way to exploit this for ill, but I haven't spent a ton of time trying either.

jeremycline:
> Ah, OK I get it now, thanks. We could use the same option and try each method and see which one works first, but I wonder it that's wise, also from a security point of view.

Yeah, that's one thing that made me shy about going this route. I don't know why anyone would have a Python script called "fedora_messaging.callbacks" in their home directory that deletes all their files in a loop at the module level, but I just *know* someone would do this and accidentally their home directory or something.

> I don't have a better idea either. Let's add a new option, and if we find the perfect way later we can deprecate it and have the code follow to the improved --callback codepath.

Sounds good to me, assuming Patrick doesn't object.

fedora_messaging/cli.py
puiterwijk: So, I will say that this whole idea gives me an unhappy feeling, I do get the use.
I think that any way we add that lets you import arbitrary code as callback will be equally dangerous: still the same level of access required to register e.g. setuptools entrypoints or other packages.
Given that this needs to be called explicitly on the CLI, I guess it's as good as any arbitrary callback mechanism we're going to get.

So +1 for the `# nosec` here as long as this function only ever gets called with values passed in explicit CLI arguments (as it is right now).

puiterwijk: So explicitly, it must *not* be called based on contents of received messages.

fedora_messaging/tests/fixtures/callback.py
pypingou: missing a docstring with the source :laughing:

jeremycline:
@abompard, since Patrick approves, are you okay with this?

Issue #155, opened 2019-04-12
trigger tool


nirik:
fedmsg has a tool called 'fedmsg-trigger'. It lets you easily do a one command line type thing to run a command/script when a specific message comes accross the bus. 

It would be cool if there was something like this for fedora-messaging too.

abompard:
It is indeed an interesting feature. It can be done with the current code, although not as easily as the `fedmsg-trigger` command. You'll need to define a callback module like this (for example called `run_command.py`):
```python
from subprocess import call
from fedora_messaging.config import conf
def run(message):
    call(conf["consumer_config"]["command"], shell=True)
```
And a configuration file that contains these lines (named for example `trigger.toml`):
```toml
# The module and function name
callback = "run_command:run"
[consumer_config]
command = "echo message received"
```
The module must be importable, so you may need to set `PYTHONPATH`, but then you can run the command `PYTHONPATH=. fedora-messaging --conf trigger.toml consume` and it will execute the command on each received message. You can filter the topics you want to act on by configuring the `[[bindings]]` section in the configuration file (the `routing_keys` list is topics or topic patterns).

Would that let you do what you want to do?

jeremycline:
@abompard, I was thinking we could add a couple generally useful callbacks to a ``callbacks.py`` module (and move the printer out into that) that we document and maintain. That way folks can just write some configuration and not have to deal with writing Python.

It's only tangentially related, but I wonder if we should also allow specifying a Python file instead of only a Python path so folks don't need to deal with packaging.

abompard:
Yeah I agree to both propositions.

PR #163, opened 2019-04-17
WIP: Start on a set of common callbacks

jeremycline:
Move the printer callback into the "callbacks" module and add the callback requested in #155.

There are no tests yet because I wanted to get a bit of feedback on usage. @nirik, would the command ever need to see the message itself, or is this purely a way to trigger a command based on the topic alone? Would it ever be helpful to requeue the message and run the command again if it gets stuck while running or exits with some non-zero return code?

The current configuration options are:
* ``command`` (string or list of strings): The arguments to launch the process
        as a string or list of strings.
* ``shell`` (boolean): Whether or not to execute the command in a shell. Refer to the
        `security considerations
        <https://docs.python.org/3/library/subprocess.html#security-considerations>`_
        before setting this to true. Defaults to false.
* ``timeout`` (integer): The length of time, in seconds, to wait on the command
        before halting the command and returning the message to the queue for later
        processing. Defaults to no timeout.

Would anything else be useful to configure here?

### DIFF
--- a/fedora_messaging/tests/unit/test_callbacks.py
+++ b/fedora_messaging/tests/unit/test_callbacks.py
@@ -1,5 +1,5 @@
 # This file is part of fedora_messaging.
-# Copyright (C) 2018 Red Hat, Inc.
+# Copyright (C) 2020 Red Hat, Inc.
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


cverna:
I think we should go with that for now, and based on the usage ask people to file RFEs. I have added a simple test case for the runner callback.

Do we need anything else to merge this PR ?

fedora_messaging/tests/unit/test_callbacks.py
```diff
@@ -48,6 +49,18 @@ def test_printer(self):
...
+class RunnerTests(unittest.TestCase):
+    def test_runner(self):
+        """ Assert the run callback calls a subprocess."""
+        config.conf = config.LazyConfig()
+        config.conf["consumer_config"]["command"] = "/bin/true"
+        message = api.Message(body="Hello world", topic="hi")
+        with mock.patch("subprocess.check_call") as mock_subprocess:
+            callbacks.run(message)
+
+        mock_subprocess.assert_called_once_with("/bin/true", shell=False, timeout=None)
```
abompard: Could you also check that the `Drop` exception is raised when the command fails, and that the proper exception is raised when the command times out?

jeremycline:
Seems likely I'm not going to finish this.

PR #177, opened 2019-05-20
Re-organize the consuming documentation for clarity

jeremycline:
A number of users did not find the section on using a class as a
callback. This attempts to re-organize the documentation around the
consumer API to make it clearer.

@dustymabe, @bowlofeggs, it'd be great if you could skim this and let me know if it's clearer.

### DIFF
--- a/docs/consuming.rst
+++ b/docs/consuming.rst
@@ -79,39 +79,89 @@ Consult the manual page for complete details on this command line interface.
 .. note:: For users of fedmsg, this is roughly equivalent to fedmsg-hub
 
 
-systemd Service
-===============
+Consumer API
+============
 
-A systemd service file is also included in the Python package for your
-convenience. It is called ``fm-consumer@.service`` and simply runs
-``fedora-messaging consume`` with a configuration file from
-``/etc/fedora-messaging/`` that matches the service name::
+The introduction contains a very minimal callback. This section covers the
+complete API for consumers.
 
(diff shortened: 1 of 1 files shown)


dustymabe:
I think it would have helped me today if we add an example of calling that example class via the CLI. i.e. `fedora-messaging --conf /work/my_config.toml consume --callback=modname:SaveMessage`

docs/consuming.rst
```diff
@@ -79,39 +79,84 @@ Consult the manual page for complete details on this command line interface.
...
+The introduction contains a very minimal callback. This section covers the
+complete API for consumers.
 
-    $ systemctl start fm-consumer@sample.service  # uses /etc/fedora-messaging/sample.toml
+The Callback
+------------
+
+The callback provided to :func:`fedora_messaging.api.consume` or the command-line
+interface can be any callable Python object, so long as it accepts the message
+object as a single position argument.
```
bowlofeggs:
```suggestion
object as a single positional argument.
```

jeremycline: Fixed, thanks!

jeremycline:
> I think it would have helped me today if we add an example of calling that example class via the CLI. i.e. fedora-messaging --conf /work/my_config.toml consume --callback=modname:SaveMessage

I've added a note following the example class.

dustymabe:
> > I think it would have helped me today if we add an example of calling that example class via the CLI. i.e. fedora-messaging --conf /work/my_config.toml consume --callback=modname:SaveMessage
> 
> I've added a note following the example class.

👍

PR #178, opened 2019-05-21
Reset the connection and channel for AMQP errors

jeremycline:
If a generic AMQP error occurs, do not try to re-use the connection. The
channel and connection has likely been closed because this is usually
due to protocol-level issues.

### DIFF
--- a/fedora_messaging/_session.py
+++ b/fedora_messaging/_session.py
@@ -167,10 +167,13 @@ class PublisherSession(object):
                 _log.error(str(e))
                 if self._connection and self._connection.is_open:
                     self._connection.close()
+                self._connection = self._channel = None
                 raise ConnectionException(reason=e)
         except pika_errs.AMQPError as e:
+            _log.info("Resetting connection to %s", self._parameters.host)
             if self._connection and self._connection.is_open:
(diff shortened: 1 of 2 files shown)


fedora_messaging/_session.py
```diff
@@ -167,10 +167,13 @@ def publish(self, message, exchange=None):
                 _log.error(str(e))
                 if self._connection and self._connection.is_open:
                     self._connection.close()
+                self._connection = self._channel = None
                 raise ConnectionException(reason=e)
         except pika_errs.AMQPError as e:
+            _log.info("Resetting connection to %s", self._parameters.host)
             if self._connection and self._connection.is_open:
                 self._connection.close()
+            self._connection = self._channel = None
```
bowlofeggs: The code here looks similar to the code above - would it be helpful to have a _reset_connection() to consolidate?

jeremycline: Yeah, but I'd rather get rid of this entire module in the medium term and run everything in Twisted with https://crochet.readthedocs.io/en/stable/. Dealing with AMQP in an asynchronous fashion is much simpler.

PR #182, opened 2019-06-07
Add the "conf" and "DEFAULTS" variables to the API documentation

jeremycline:
The config docs mention these, but they have not be officially
documented as part of the API. Include them.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -124,6 +124,22 @@ Exceptions
 
 .. _twisted-api:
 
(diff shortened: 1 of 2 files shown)


abompard:
The CI failure comes from Twisted 19.2.1 which [has dropped support for Python 3.4](https://github.com/twisted/twisted/blob/trunk/NEWS.rst#twisted-1920-2019-04-07)

PR #183, opened 2019-06-08
Twisted 19.2.0 is the last version to support Python 3.4

abompard:
See https://github.com/twisted/twisted/blob/trunk/NEWS.rst#twisted-1920-2019-04-07

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,6 +6,7 @@ passenv = CI TRAVIS TRAVIS_*
 deps =
     pika0x: pika<1.0.0b1
     pika1: git+https://github.com/pika/pika.git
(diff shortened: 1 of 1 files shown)


jeremycline:
So there's a minor issue here about duplicate requirements for twisted (one in the requirements file and the explicit one here), but there's a second issue I noticed: pip is also dropping 3.4 support.

The question is, do we need to support 3.4 at all? From what I understand EL7 is now Python 3.6, and we can support the 2.7 stack on EL7 or 3.6 and a virtualenv with pip. So maybe instead we should drop Python 3.4 and 3.5 from our CI.

abompard:
Yeah I agree.

PR #185, opened 2019-06-18
Drop official Python 3.4 and 3.5 support

jeremycline:
Drop Python 3.4 and 3.5 from the CI configuration and mark Python 3.6 as
the lowest Python 3 version supported. EPEL7 has a Python 3.6 stack now.

This unfortunately expanded a bit because the CI was also broken due to Travis moving to Xenial by default and not installing RabbitMQ on those *and* because Six moved its docs. I went ahead and also dropped the Pika 0.x matrix because 1.0+ is available everywhere we care about.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,38 +30,26 @@ env:
 matrix:
   include:
     - python: "2.7"
-      env: TOXENV=py27-pika1-unittest
+      env: TOXENV=py27-unittest
     - python: "3.6"
-      env: TOXENV=py36-pika1-unittest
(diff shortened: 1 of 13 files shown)


abompard:
I'm guessing the coverage went down because we still have switches in the code for Pika < 1.0. I can remove those if you want.

jeremycline:
Okay I ripped out all the pika 0.x code I could find. I'm not concerned about the coverage drop in _session since there's a PR to delete it, and in that same PR I plan to get the coverage of protocol.py up to 100%.

PR #188, opened 2019-06-20
Update example config: ZMQ exchange, extra properties, logging

AdamWill:
This is based on my experience porting several fedmsg consumers
to fedora-messaging. @jcline thought it would be good for
consumers to provide more useful info in client_properties, so
I invented a couple of optional fields there. Also you need a
second binding to receive messages from the ZMQ->AMQP bridge
(which most consumers are likely to want to do), and it seems
useful to note that if your consumer implements a logger, it
needs to be configured here or else messages won't show up.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/configs/fedora.toml
+++ b/configs/fedora.toml
@@ -12,6 +12,12 @@ certfile = "/etc/fedora-messaging/fedora-cert.pem"
 
 [client_properties]
 app = "Example Application"
+# Some suggested extra fields:
+# URL of the project that provides this consumer
+app_url = "https://github.com/fedora-infra/fedora-messaging"
+# Contact emails for the maintainer(s) of the consumer - in case the
+# broker admin needs to contact them, for e.g.
+app_contacts_email = ["jcline@fedoraproject.org"]
 
 [exchanges."amq.topic"]
 type = "topic"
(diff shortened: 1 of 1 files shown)


configs/fedora.toml
```diff
@@ -38,6 +44,15 @@ queue = "00000000-0000-0000-0000-000000000000"
 exchange = "amq.topic"
 routing_keys = ["#"]  # Set this to the specific topics you are interested in.
 
+# To receive messages from the fedmsg->fedora-messaging bridge,
+# duplicate the binding but for the 'zmq.topic' exchange. It's safe to
+# bind to all the routing_keys you are interested in on both exchanges
+# so you keep working when a publisher is ported to fedora-messaging.
```
jeremycline: Since this is only relevant to internal /pubsub users, maybe at the bottom add
```
# This is only required for users of the "/pubsub" exchange, not those using the 
# "/public_pubsub" virtual host and the "fedora" user
```

AdamWill: oh, if that's the case, I think we should leave it out; I wasn't intending to include stuff that was specific to that case here. It would be good to cover that kinda stuff somewhere, but I'm not sure where.

AdamWill:
Oh, OK, I was about to add a separate sample config file, but that works too.

Perhaps the general fedora-messaging docs should have a little pointer to that guide for people implementing consumers in Fedora infra? I never knew that guide existed...

jeremycline:
Yeah that probably wouldn't hurt.

Issue #190, opened 2019-06-22
Understanding how to connect to the public Fedora broker from outside using library


changlinli:
Using `fedora-messaging --conf=conf.toml consume --callback=fedora_messaging.example:printer` I can successfully connect and see messages printing out.

Unfortunately when I run what I think is the equivalent code from the docs, I get nothing printed out to stdout at all (not even any connection information).

```python
from fedora_messaging import api

# First, define a function to be used as our callback. This will be called
# whenever a message is received from the server.
def printer_callback(message):
    """
    Print the message to standard output.

    Args:
        message (fedora_messaging.message.Message): The message we received
            from the queue.
    """
    print(str(message))

# Next, we need a queue to consume messages from. We can define
# the queue and binding configurations in these dictionaries:
queues = {
    '00000000-0000-0000-0000-000000000000': {
        'durable': False,  # Delete the queue on broker restart
        'auto_delete': True,  # Delete the queue when the client terminates
        'exclusive': False,  # Allow multiple simultaneous consumers
        'arguments': {},
    },
}
binding = {
    'exchange': 'amq.topic',  # The AMQP exchange to bind our queue to
    'queue': '00000000-0000-0000-0000-000000000000',  # The unique name of our queue on the AMQP broker
    'routing_keys': ['#'],  # The topics that should be delivered to the queue
}

# Start consuming messages using our callback. This call will block until
# a KeyboardInterrupt is raised, or the process receives a SIGINT or SIGTERM
# signal.
api.consume(printer_callback, bindings=binding, queues=queues)
```

This is even if I set the `FEDORA_MESSAGING_CONF` environment variable to `conf.toml`. Is there some other configuration or logging that I'm missing?

praiskup:
See #171 and #186.  The example from docs doesn't work.  Well I spent quite some time on trying to make this work (without success) ... and seems like I'm not alone.

I'd suggest at least to turn on the debugging output by default in the example, and while on it - consider adding a warning that it currently doesn't work.

jeremycline:
It would be good to add the log configuration setup to the snippet, yes.

changlinli:
@jeremycline After #192 I now get a clearer idea of what's wrong (this also saves me the effort of opening another ticket for trying to connect via Java's RabbitMQ library) so thanks a bunch!

I'm now seeing

```
[INFO fedora_messaging._session] Authenticating with server using x509 (certfile: fedora-cert.pem, keyfile: fedora-key.pem)
[...]

PR #192, opened 2019-06-24
Document a quick way to setup logging

jeremycline:
Some users are hitting issues and aren't aware that logging needs to be
configured. Add a note to the consuming snippets about setting up
logging.

See #190 and the issues mentioned in the comment.


praiskup:
thank you

PR #193, opened 2019-06-26
Correct queue name in consumer example

changlinli:
The queue name has to be a UUID otherwise the server errors out with a
permission issue.

### DIFF
--- a/docs/consuming.rst
+++ b/docs/consuming.rst
@@ -46,7 +46,7 @@ argument::
     # Next, we need a queue to consume messages from. We can define
     # the queue and binding configurations in these dictionaries:
     queues = {
-        'demo': {
+        '00000000-0000-0000-0000-000000000000': {
             'durable': False,  # Delete the queue on broker restart
             'auto_delete': True,  # Delete the queue when the client terminates
             'exclusive': False,  # Allow multiple simultaneous consumers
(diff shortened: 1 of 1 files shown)


abompard:
That's only true for the Fedora public broker, internal queue names are less constrained, but it's a good point, thanks.

jeremycline:
This is documented at https://fedora-messaging.readthedocs.io/en/stable/fedora-broker.html. I wonder if it would be better to have a "Getting Started" section before the details of publishing or consuming that covers:

1. How to set up a local environment in Fedora (broad strokes, just inform users about rabbitmq and perhaps the management console).
2. How to consume from the Fedora public vhost.
3. How to consume from and publish to the Fedora private vhost

changlinli:
Yep that might do the trick. The central problem right now is that if users try to use that snippet of code to connect to the public Fedora infrastructure, they'll get an access error that might draw their attention away from the fact that the queue name is incorrect. FWIW I found where I had messed up from that documentation, but at the time there was another auth problem that seems to have gone away with the latest release of `fedora-messaging` that confounded some things.

```
[WARNING pika.channel] Received remote Channel.Close (403): "ACCESS_REFUSED - access to queue 'demo' in vhost '/public_pubsub' refused for user 'fedora'" on <Channel number=1 OPEN conn=<SelectConnection OPEN transport=<pika.adapters.utils.io_services_utils._AsyncSSLTransport object at 0x7f0ac4849320> params=<URLParameters host=rabbitmq.fedoraproject.org port=5671 virtual_host=/public_pubsub ssl=True>>>
```

Basically there's no ready-made "Hello World"-esque snippet of code that works for connecting to the Fedora public infrastructure like the CLI has.

Happy to close out this PR in favor of something that has those three options.

Otherwise another option is to instead of changing `demo`, just make a note that it needs to be a UUID in a comment above that line if you want to connect to the public vhost.

changlinli:
Closing in favor of #196

PR #194, opened 2019-06-26
Drop the deprecated Twisted APIs

jeremycline:
The original Twisted APIs were replaced with "fedora_messaging.api.twisted_consume". Remove the interface documentation and drop the code. This is backwards-incompatible with the 1.x series, but it's very unlikely anyone used this code and even if they did they have had a reasonable time when the interface was deprecated.

Note: this is based on top of the crochet work so that needs merging first. I'm opening this now to see the combined test coverage.


jeremycline:
I'm fine just leaving these in for now since I don't have a good reason to break the APIs. They're documented as deprecated so people should stay away from them.

PR #196, opened 2019-06-27
docs: create a quick-start guide

jeremycline:
Provide users with a quick-start guide that covers local brokers for
development purposes, connecting to the public Fedora broker, and a
pointer to the infrastructure development guide for getting connected to
the internal broker.

@changlinli


changlinli:
Love it! Thanks so much!

PR #198, opened 2019-07-23
Fix exception arguments and documentation

hluk:
Signed-off-by: Lukas Holecek <hluk@email.cz>

### DIFF
--- a/fedora_messaging/exceptions.py
+++ b/fedora_messaging/exceptions.py
@@ -71,15 +71,28 @@ class ConfigurationException(BaseException):
         self.message = message
 
     def __str__(self):
-        return "Configuration error: " + self.message
+        return "Configuration error: {}".format(self.message)
 
 
 class PublishException(BaseException):
-    """Base class for exceptions related to publishing."""
+    """
+    Base class for exceptions related to publishing.
+
+    Args:
+        reason (str): The reason the server gave for the publish error.
+    """
 
     def __init__(self, reason=None, **kwargs):
         super(PublishException, self).__init__(**kwargs)
         self.reason = reason
+
+    def __str__(self):
+        if self.reason:
+            return "Publish error: {}".format(self.reason)
+        return "Publish error"
+
+    def __repr__(self):
+        return "PublishException(reason={})".format(self.reason)
 
 
 class PublishReturned(PublishException):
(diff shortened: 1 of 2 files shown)


fedora_messaging/exceptions.py
```diff
@@ -17,7 +17,7 @@ class PermissionException(BaseException):
         obj_type (str): The type of object being accessed that caused the
             permission error. May be None if the cause is unknown.
         description (object): The description of the object, if any. May be None.
-        reason (str): The reason the server gave for the permission error, if
+        reason (str): The reason the server gave the permission error, if
```
pypingou: The `for` seems justified for me, it's the reason for the error, no?

pypingou:
Looks good to me, though I expect we may want tests with this?

fedora_messaging/exceptions.py
hluk: I misread it. Sometimes I don't understand English. I'm reverting it.

hluk:
Any idea what's up with the integration test? For me they get also stuck locally.

fedora_messaging/exceptions.py
```diff
@@ -75,12 +75,25 @@ def __str__(self):
...
+        reason (str): The reason the server gave for the publish error.
+    """
 
     def __init__(self, reason=None, **kwargs):
         super(PublishException, self).__init__(**kwargs)
         self.reason = reason
 
+    def __str__(self):
+        if self.reason:
+            return "Publish error: " + self.reason
```
jeremycline: It'd be good to call str on self.reason so users can pass in non-text objects (for example, a lower-level exception)

fedora_messaging/exceptions.py
```diff
@@ -106,11 +119,22 @@ class ConnectionException(BaseException):
...
-    def __init__(self, *args, **kwargs):
-        super(ConnectionException, self).__init__(*args)
-        self.reason = kwargs.get("reason")
+    def __init__(self, reason=None, **kwargs):
+        super(ConnectionException, self).__init__(**kwargs)
+        self.reason = reason
+
+    def __str__(self):
+        if self.reason:
+            return "Connection error: " + self.reason
```
jeremycline: Same deal here with str

jeremycline:
Actually, current master also passes CI. Maybe one of the tests is expecting something from one of these exceptions that broke? I don't have time to dig into it today, but I might be able to in the next couple days.

PR #209, opened 2019-10-21
Fix a reconnection issue

abompard:
Reconnecting twice in a row would trigger an `AlreadyCalledError` on the `_client_deferred` Deferred.

### DIFF
--- a/fedora_messaging/twisted/factory.py
+++ b/fedora_messaging/twisted/factory.py
@@ -311,6 +311,8 @@ class FedoraMessagingFactoryV2(protocol.ReconnectingClientFactory):
             self.resetDelay()
             self._client = client
             self._client_deferred.callback(client)
+            # Renew the deferred to handle reconnections.
+            self._client_deferred = defer.Deferred()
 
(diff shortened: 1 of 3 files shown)


fedora_messaging/twisted/factory.py
```diff
@@ -310,7 +310,9 @@ def on_ready(unused_param=None):
             _std_log.debug("AMQP handshake completed; connection ready for use")
             self.resetDelay()
             self._client = client
-            self._client_deferred.callback(client)
+            if not self._client_deferred.called:
```
jeremycline: This doesn't feel like quite the right approach because we've just made a new client and are keeping around a deferred object that presumably fired with the old client, right?

Is the right approach to do something like:
```
self._client_deferred.callback(client)
self._client_deferred = defer.Deferred()
```

And then ``when_connected`` starts handing out the fresh deferred only if ``self._client and not self._client.is_closed`` is untrue.

I'll admit to being uncertain about this, I've not looked at this code in a while.

abompard: Yeah that seems much better. I've updated the PR.

Issue #212, opened 2019-12-12
AttributeError: 'NoneType' object has no attribute 'publish'


Zlopez:
Plenty of test started failing on Anitya with fedora-messaging 2.0.0 with the error in title.

Here is the link for the travis job, so you can look at the logs: https://travis-ci.org/release-monitoring/anitya/builds/624136745

Locally I have fedora-messaging 1.7.2 and this issue didn't happened.

abompard:
Thanks for reporting, I've opened PR #213 to fix the error.

However, a lot of your unit test code is currently not calling `fedora_messaging.testing.mock_send()`, therefore your tests are delayed because Fedora Messaging tries to send messages to a non-existant RabbitMQ, and tries to reconnect periodically until it hits the publishing timeout. I'll send you a PR to fix that.

Zlopez:
I know that some tests sending fedora messages are not calling `mock_send()`. I should've fix this some time ago, but didn't found time. :-(

PR #213, opened 2019-12-13
Fix handling of new connections after a publish timeout

abompard:
When a `PublishTimeout` occurs, the `_client_deferred` is cancelled but it was not renewed, as it is in other cases.

Fixes: #212


abompard:
Thanks Jeremy, I'll cut a new release because this is impacting quite a few apps these days (there seem to be connectivity issues with the rabbitmq cluster in infra).

PR #214, opened 2020-01-06
Only try to restart fm-consumer@ services every 60 seconds

AdamWill:
Consumers using the fm-consumer@ service pattern are quite
vulnerable to a problem where, for instance, consuming a message
fails because some key service the consumer needs to talk to is
down. When this happens, the service is considered to have
failed, and - because Restart=on-failure is set - is immediately
(well, after 100ms) restarted. It then immediately attempts to
consume the same message, and - because the service likely
didn't come up in the last 100ms - probably fails again. This
continues to happen in a very fast cycle. If the consumer times
out quite quickly and the service stays down for several hours,
you can easily accumulate a loop of tens of thousands of
failures. This pollutes the system journal quite badly and is
a nightmare if you have the consumer configured to email the
administrator on failure; the admin's email will get tens of
thousands of failure mails.

Some kind of incremental backoff would be ideal here, but sadly
systemd doesn't have that implemented yet:
https://github.com/systemd/systemd/issues/6129
in its absence, this just sets RestartSec=60, meaning the
service will only be restarted every 60 seconds. This should
cut down the floods quite a lot.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/fm-consumer@.service
+++ b/fm-consumer@.service
@@ -6,6 +6,7 @@ Documentation=http://fedora-messaging.readthedocs.io/
 Type=simple
 ExecStart=/usr/bin/fedora-messaging --conf /etc/fedora-messaging/%i.toml consume
 Restart=on-failure
+RestartSec=60
 
 [Install]
 WantedBy=multi-user.target


AdamWill:
@lruzicka for our sanity ;)

abompard:
Thanks for the PR Adam. The key service you're talking about here is not the broker, right? Then shouldn't it be the callback's job to wait for the service to come up? Why would it result in the consumer being considered failed and restarted?
I think I'd understand better with an example, if you have that available.

AdamWill:
By the 'callback' you mean the actual consumer class, right? Here's an example of actual consumers I have this issue with:

https://pagure.io/fedora-qa/fedora_openqa/blob/master/f/fedora_openqa/consumer.py#_167

`OpenQAWikiReporter` and `OpenQAResultsDBReporter` are consumer classes that take an 'openQA job done' message and file a corresponding report to (respectively) the wiki and [ResultsDB](https://taskotron.fedoraproject.org/resultsdb/results). They're run by the `fm-consumer@` pattern, with a config to listen for `org.fedoraproject.prod.openqa.job.done` messages, like this:

    [[bindings]]
    queue = "openqa_resultsdb_reporter"
    exchange = "amq.topic"
    routing_keys = ["org.fedoraproject.prod.openqa.job.done",]

if the wiki or resultsdb happen to be down, the attempt to send the report to them fails, and this results in the service being marked as failed, because the message wasn't successfully consumed. It then immediately restarts and tries to send the report again. Forever.

There was an especially bad case of this the other day because resultsdb was down, but connections to it weren't timing out - they were immediately returning a 503 error. So we were getting, like, one retry per second. For several hours.

Are you saying I should handle more errors in the consumer class itself? Say, if it can't reach the server, do a backoff / wait / retry cycle itself? I mean, I could do that, I guess, I just didn't expect that was how it was intended to work.

jeremycline:
I didn't expect folks to explicitly not handle exceptions in consumers. It's true the consumer will clean up the message state and halt, but that's not really meant to be how you handle expected failures. Of course, if you'd prefer to rely on systemd for your backoff/retry mechanisms, I suppose it works. Still, everyone has different backoff requirements so I don't really encourage the pattern.

AdamWill:
I mean, it's not an 'expected failure', exactly. The key point is: when the service comes back up, we need to handle the message. We don't want to just drop it and never file the result if the service happens to be down when the message comes in. So I can't just say 'oh well, connection failed, never mind, we've handled this message!', right? What is a consumer supposed to do in that situation?

jeremycline:
> I mean, it's not an 'expected failure', exactly. The key point is: when the service comes back up, we need to handle the message. We don't want to just drop it and never file the result if the service happens to be down when the message comes in. So I can't just say 'oh well, connection failed, never mind, we've handled this message!', right? What is a consumer supposed to do in that situation? Basically just keep blocking until it can get through?

Right, you should be free to block as long as you need to. That being said, there's not a huge downside to crashing and you can override systemd units easily enough so I guess it's a purely academic discussion.

AdamWill:
Oh, heh, I was really just asking questions to clarify what I should do instead of this - wasn't expecting you to merge it. But hey, it still shouldn't hurt anything :)

jeremycline:
I think it's a solid change regardless because I expect plenty of people will take the same approach and in those cases we don't want to hammer the message broker with tons of short-lived connections.

abompard:
> I think it's a solid change regardless because I expect plenty of people will take the same approach and in those cases we don't want to hammer the message broker with tons of short-lived connections.

Ah that's a good point.

PR #219, opened 2020-06-30
Update the schema docs

abompard:
Update some documentation on the message schemas.


nphilipp:
Looks like I'll work on fedora_messaging in the near future, so I might as well get my feet wet. :wink:

docs/sample_schema_package/mailman_messages/messages.py
```diff
@@ -69,7 +76,10 @@ def url(self):
 
     @property
     def app_icon(self):
-        """An URL to the icon of the application that generated the message."""
+        """An URL to the icon of the application that generated the message.
```
nphilipp, line 79: It should be "A URL...", it's only "an" if followed by a spoken vowel.

docs/sample_schema_package/mailman_messages/messages.py
```diff
@@ -82,6 +92,13 @@ def packages(self):
         """List of packages affected by the action that generated this message."""
         return []
 
+    def _get_avatar_from_from_header(self, from_header):
+        """Converts a From email header to an avatar."""
+        # Extract the username
+        addr = parseaddr(from_header)[1]
+        username = addr.split("@")[0]
+        return schema_utils.user_avatar_url(username)
+
```
nphilipp, line 101: Hmm, previously it would have looked up an avatar by the full email address, now only by the username. Has anything changed underneath that this is fixing?

docs/tutorial/schemas.rst
```diff
@@ -21,10 +21,18 @@ Copy the ``docs/sample_schema_package/`` directory from the
 ``fedora-messaging`` git clone to your app directory.
 
 Edit the ``setup.py`` file to change the package metadata. Rename the
-``mailman_schema`` directory to something relevant to your app, like
-``yourapp_message_schemas``. There is no naming convention at the moment.
+``mailman_messages`` directory to something relevant to your app, like
+``yourapp_message_messages``. This is the naming convention.
```
nphilipp, line 25: I guess this should be "yourapp_messages" instead.

docs/tutorial/schemas.rst
```diff
@@ -159,7 +167,7 @@ however you named the package) to construct your message instances and call
 
     from fedora_messaging.api import publish
     from fedora_messaging.config import conf
-    from yourapp_message_schema.schema import Message
+    from yourapp_message_messages.messages import Message
```
nphilipp, line 170: Likewide here ("yourapp_messages.messages").

docs/tutorial/schemas.rst
```diff
@@ -197,8 +205,8 @@ receivers, restart them, and then start using the new version in the publishers.
 You should keep the old schema versions in the schemas package for a reasonable
 amount of time, long enough to make sure all receivers are up-to-date. To avoid
 clutter, we recommend you use a separate module per schema version
-(``yourapp_message_schemas.v1:Message``,
-``yourapp_message_schemas.v2:Message``, etc)
+(``yourapp_message_messages.v1:Message``,
+``yourapp_message_messages.v2:Message``, etc)
```
nphilipp, line 209: And here (I guess "yourapp_messages.v1:Message", "yourapp_messages.v2:Message").

docs/sample_schema_package/mailman_messages/messages.py
abompard: Yes, the `user_avatar_url` function is expecting usernames, not email adresses, the previous call was wrong.

abompard:
Thanks for the review Nils, my search-and-replace did some pretty bad damage apparently... It's fixed now.

docs/messages.rst
```diff
@@ -84,19 +84,19 @@ Packaging
 
 Finally, you must distribute your schema to clients. It is recommended that you
 maintain your message schema in your application's git repository in a separate
-Python package. The package name should be ``<your-app-name>_schema``.
+Python package. The package name should be ``<your-app-name>_messages``.
```
praiskup, line 87: Can we please suggest `-` instead of `_`?  Package names with underscores look ugly :-)

praiskup: FWIW, we used `python3-copr-messaging` name, but we can add the provide (and in future split the package into two parts).

abompard: I agree it looks ugly, but I think there is (or was?) a bug in pip that made it treat differenly packages with a dash and with an underscore. I'll check that and see if it's still valid.

Don't worry about renaming, it's not a hard convention, just a suggestion for new packages.

praiskup: > I agree it looks ugly, but I think there is (or was?) a bug in pip that made it treat differenly packages with a dash and with an underscore. I'll check that and see if it's still valid.

The problem I'm only aware of is that a python module name can not contain
dashes -- so you can `import some.thing.some_where`.

But the package (project) name and the python module name don't have to be
the same.

docs/conf.py
```diff
@@ -17,7 +17,7 @@
 
 sys.path.insert(0, os.path.abspath("../"))  # NOQA
 
-import fedora_messaging
+import fedora_messaging  # NOQA
```
nphilipp, line 20: We could even make it `# noqa: F401` to only ignore the unused import (assuming this is what it's about).

abompard: It's about the import not being at the top of the file. I'll be more specific, thanks.

PR #235, opened 2021-05-10
Support Forbidden errors on publish

abompard:
When [topic authorization](https://www.rabbitmq.com/access-control.html#topic-authorisation) is enabled, the broken can return Forbidden messages when messages are published to a forbidden topic. Handle this type of response instead of retrying endlessly.

### DIFF
--- a/docs/tutorial/conversion.rst
+++ b/docs/tutorial/conversion.rst
@@ -117,7 +117,7 @@ to encapsulate the messages. The wrapper could look like::
 
     from fedora_elections_messages.schema import Message
     from fedora_messaging.api import publish as fm_publish
-    from fedora_messaging.exceptions import PublishReturned, ConnectionException
+    from fedora_messaging.exceptions import PublishReturned, PublishForbidden, ConnectionException
 
     LOGGER = logging.getLogger(__name__)
 
@@ -127,7 +127,7 @@ to encapsulate the messages. The wrapper could look like::
                 topic="fedora.elections." + topic,
                 body=msg,
             ))
-        except PublishReturned as e:
+        except (PublishReturned, PublishForbidden) as e:
             LOGGER.warning(
                 "Fedora Messaging broker rejected message %s: %s",
                 msg.id, e
@@ -235,7 +235,7 @@ Now you can replace the current call to fedmsg with a call to
 
     import logging
     from fedora_messaging.api import Message, publish
-    from fedora_messaging.exceptions import PublishReturned, ConnectionException
+    from fedora_messaging.exceptions import PublishReturned, PublishForbidden, ConnectionException
 
     LOGGER = logging.getLogger(__name__)
 
@@ -247,7 +247,7 @@ And replace the call to ``fedmsg.publish`` with::
             body=payload,
         )
         publish(msg)
-    except PublishReturned as e:
+    except (PublishReturned, PublishForbidden) as e:
         LOGGER.warning(
             "Fedora Messaging broker rejected message %s: %s",
             msg.id, e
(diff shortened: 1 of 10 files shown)


fedora_messaging/cli.py
```diff
@@ -383,7 +383,7 @@ def publish(exchange, file):
             click.echo("Publishing message with topic {}".format(msg.topic))
             try:
                 api.publish(msg, exchange)
-            except exceptions.PublishReturned as e:
+            except (exceptions.PublishReturned, exceptions.PublishForbidden) as e:
```
jeremycline, line 386: I'd except the generic PublishException here since all it's going to do is report the error and doesn't differentiate on the type.

fedora_messaging/api.py
```diff
@@ -298,6 +298,8 @@ def publish(message, exchange=None, timeout=30):
             message.
         fedora_messaging.exceptions.PublishTimeout: Raised if the broker could not be
             contacted in the given timeout time.
+        fedora_messaging.exceptions.PublishForbidden: Raised if the broker rejects the
+            message because of permission issues.
```
jeremycline, line 302: I think we should document that users should *always* include the ``except fedora_messaging.exceptions.PublishException`` block in error handling for this call. I can imagine users might not have that and expanding the exception list is an API change (although obviously it breaks in a much worse way without this).

Perhaps something like this in the Raises section:
`fedora_messaging.exceptions.PublishException: The parent class of PublishReturned, PublishTimeout, and PublishForbidden. Any new exceptions raised by this function will sub-class this and it should be handled for compatibility with future releases of fedora-messaging.`

docs/tutorial/conversion.rst
```diff
@@ -127,7 +127,7 @@ to encapsulate the messages. The wrapper could look like::
                 topic="fedora.elections." + topic,
                 body=msg,
             ))
-        except PublishReturned as e:
+        except (PublishReturned, PublishForbidden) as e:
```
jeremycline, line 130: See the notes below about documenting the parent class being raised, it might be good to add a generic "PublishException" block to these code snippets

jeremycline (commented):
Looks good, much better than spinning forever 😄 

I left some notes for API docs so that hopefully no one gets bit by new exceptions, but otherwise 👍🏼

fedora_messaging/api.py
abompard: Yeah, that was my first try but then I thought that `PublishTimeout` could be handled differently, because users may want to retry in this case. It's not a permanent error. What do you think would be the best way to handle that difference?

docs/tutorial/conversion.rst
abompard: Same comment about permanent vs temporary errors

fedora_messaging/cli.py
abompard: Yeah but `PublishTimeout` is handled differently just under it, and it's also a subclass of `PublishException`.

jeremycline: Woops, meant to comment *and* resolve and now GitHub ate my comment.

Anyway, I didn't notice the block below, so yeah this suggestion isn't reasonable, and there's a general "PublishException" block after the timeout case anyway so we're good.

fedora_messaging/api.py
jeremycline: I'm not sure the best way to document it, but what I want to convey to the user is their error blocks should be:

```
try:
    api.publish(...)
except PublishTimeout:
    # Special timeout handling
except PublishForbidden:
    # Special forbidden handling
except PublishException:
    # General unexpected exception, either a new exception in the library or the PublishRefused
    # exception we didn't handle explicitly above
```

It's not really related to this change specifically, and I suppose the error is general enough that most applications would prefer to just halt anyway... Maybe it doesn't matter that much.

abompard: As a library user I think that what I want to know is whether I should retry publishing or just log and drop.

jeremycline: Makes sense. I'm fine with all this as-is, then.

jeremycline (approved):

PR #239, opened 2021-06-11
Support anonymous (server-named) queues

abompard:
See https://www.rabbitmq.com/queues.html#server-named-queues

This is an API-breaking change because the queue created by the CLI will not have the same properties as before. They are now mirrored on the anonymous queues: non-durable, auto-deleted and exclusive.

It is no longer necessary to declare a queue in the configuration file: a server-named queue will be created if none is defined. Configured bindings which do not specify a queue name will be applied to the server-named queue.

It's a rather large changeset, so I'm very interested in your opinions, especially from a design standpoint.

Please look at the commits individually for more clarity ;-)

### DIFF
--- a/configs/fedora.stg.toml
+++ b/configs/fedora.stg.toml
@@ -26,21 +26,26 @@ auto_delete = false
 arguments = {}
 
 # Queue names *must* be in the normal UUID format: run "uuidgen" and use the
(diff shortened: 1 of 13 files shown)


jeremycline (approved):
Looks reasonable to me, for what it's worth. Probably deserves a major version bump as some defaults changed, but it'll be an easy migration.

PR #241, opened 2021-09-29
provide isolated development environment and fix #240

### LINKED ISSUE #240: The message serialization format should not mandate the `queue`
When loading messages from the serialized format in order to send them, the "queue" attribute will not be used. It should therefore not be mandated by the JSON schema.

When the `queue` attribute is not specified or set to `None`, we get:
```
 
Error: Unable to validate message: None is not of type 'string'
 
Failed validating 'type' in schema['properties']['queue']:
    {'description': 'The queue the message arrived on, if any.',
     'type': 'string'}
 
On instance['queue']:
    None
```

onur-ozkan:
This PR fixes #240 and provides a containerized environment for development on both docker and podman.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -89,7 +89,7 @@ For example,
     towncrier create PR42.feature    
 
 The contents of the news fragment must be written in RST format.
-See the `towncrier documentation formore information
+See the `towncrier documentation for more information
 <https://github.com/twisted/towncrier>`_.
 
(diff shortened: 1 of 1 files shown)


onur-ozkan:
@abompard Can you review this ?

abompard:
Hey @ozkanonur , thanks for the contribution! Could you please separate the fix for #240 and the container environment in two separate PRs? And rebase and target the `develop` branch, while you're at it :-)
Thanks !

PR #243, opened 2021-10-26
Fixing validation exception of queue field on serialized schemas

### LINKED ISSUE #240: The message serialization format should not mandate the `queue`

onur-ozkan:
Fixes #240

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,3 +110,7 @@ config.toml
 
 # VS Code
 .vscode
+
+# Vim
+*.swp
+*.swo
--- a/fedora_messaging/message.py
+++ b/fedora_messaging/message.py
@@ -635,11 +635,11 @@ SERIALIZED_MESSAGE_SCHEMA = {
         "id": {"type": "string", "description": "The message's UUID."},
         "body": {"type": "object", "description": "The message body."},
         "queue": {
-            "type": "string",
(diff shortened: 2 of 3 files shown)


abompard:
Hey! Thanks for the commit. Could you add the Signed-Off-By tag (if you do agree with the [DCO](https://fedora-messaging.readthedocs.io/en/stable/contributing.html#licensing) of course) please?
You can do it by running `git commit --amend -s` and the `git push --force`.

onur-ozkan:
> Hey! Thanks for the commit. Could you add the Signed-Off-By tag (if you do agree with the [DCO](https://fedora-messaging.readthedocs.io/en/stable/contributing.html#licensing) of course) please? You can do it by running `git commit --amend -s` and the `git push --force`.

Of course, sorry about I forgot doing that.

abompard (approved):

PR #263, opened 2022-05-16
Correct sync list, add issue repo and version info

gridhead:
Signed-off-by: Akashdeep Dhar <akashdeep.dhar@gmail.com>

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -2,11 +2,12 @@
 # https://packit.dev/docs/configuration/
 
 specfile_path: fedora-messaging.spec
+issue_repository: https://github.com/fedora-infra/fedora-messaging
 
 # add or remove files that should be synced
 files_to_sync:
     - fedora-messaging.spec
-    - packit.yaml
+    - .packit.yaml
 
 # name in upstream package repository or registry (e.g. in PyPI)
 upstream_package_name: fedora-messaging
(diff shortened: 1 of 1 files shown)


abompard (dismissed):

gridhead:
Assuming it didn't find `git` (ref. https://github.com/fedora-infra/fedora-messaging/runs/6452182270?check_suite_focus=true#step:4:12), I added a directive to explicitly install it (ref. https://github.com/fedora-infra/fedora-messaging/pull/263/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R31). Let us see if it works.

gridhead:
Hmm. This is strange. I have git installed and I can confirm that I am in the correct directory (ref. https://github.com/fedora-infra/fedora-messaging/runs/6463365423?check_suite_focus=true). It still does not seem to work.

gridhead:
This is where I took reference from https://github.com/pre-commit/pre-commit/issues/456. I will go out there and take a walk before I have at this (again).

abompard:
Did you have a look at the pre-commit error log? I suspect it may be related to the git safe-directory feature I fixed in FASJSON: https://github.com/fedora-infra/fasjson/commit/8f13c92964d5ba70a14477a3844b11f72f5e80c6.
I'll see if I can apply the same fix here.

gridhead:
I tried to add a task to `cat` the logs after the job failed but then I remember that nothing really executes once a command exits out with a non-zero exit code. So it didn't quite work.

abompard:
Well, it's more complex than I thought, so it'll have to wait for tomorrow if you don't mind (I'm supposed to be off today ;-) ).

gridhead:
@abompard, apologies. I am totally okay with waiting on this. I will probably bang my head on Noggin in the meanwhile. :P

abompard:
Alright, please rebase, it should pass now.

abompard (approved):

PR #274, opened 2022-10-06
Use tomllib from the standard library on Python 3.11

hroncok:
The code would be slightly easier if we fell back to tomli, but I kept using toml as this code also targets EL 8.

See https://fedoraproject.org/wiki/Changes/DeprecatePythonToml

### DIFF
--- a/fedora_messaging/config.py
+++ b/fedora_messaging/config.py
@@ -285,7 +285,12 @@ import logging.config
 import os
 
 import pkg_resources
-import toml
+
+
+try:
+    import tomllib
+except ImportError:
+    import tomli as tomllib
 
 from . import exceptions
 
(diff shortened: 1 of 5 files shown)


hroncok:
FTR tomli is being added to epel8, so this could be made a bit simpler.

hroncok:
https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2022-5e36342ad8

abompard:
OK, do you want to rework it Miro?

hroncok:
Done.

hroncok:
The mocks got unhappy, trying to fix that.

hroncok:
TOMLDecodeError does not seem to have line numbers. Is that a problem?

hroncok:
This turned out to be more tedious than I anticipated. The tests are very unforgiving to slight implementation changes :(

Anyway, I think I cracked it and

    $ tox -e=py3{6,7,8,9,10,11}-unittest

Passes.

Let me know if I shall squash or if you'd rather squash when merging. I'd like you to see my train of thought when reviewing it, so not squashing yet.

abompard:
> Let me know if I shall squash or if you'd rather squash when merging. I'd like you to see my train of thought when reviewing it, so not squashing yet.

Yeah, got it, thanks. Please squash. Sorry it ended up more painful that it should have been.

hroncok:
Squashed, signed off.

PR #334, opened 2024-03-04
Added a replay command

### LINKED ISSUE #332: Add a replay command
Add a replay command that would get a message from datagrepper and replay it to the bus.

Get inspiration from https://github.com/fedora-infra/the-new-hotness/blob/master/devel/fedora_messaging_replay.py but use `message.load_message()` to preserve headers.

Freedisch:
# Description of changes
- Related issue #332 
- Added a new command replay is added to the CLI, taking a single argument message_id, which represents the ID of the message to replay.
- The publish_message function uses `Message.load_message` to create a Message object from the JSON data, ensuring all headers and properties are preserved, and then publishes it using `api.publish`.


Freedisch:
I'm not quite sure of the reason behind the failing of these CI tests, but I'm exploring it to see what's causing it🤔

gridhead:
The tests might not necessarily fail from the changes that you made as they are installability tests but try adjusting the version requirement for the `pytz` dependency. From https://github.com/fedora-infra/fedora-messaging/pull/334/checks?check_run_id=22273594026, I got to the log dumps present at https://download.copr.fedorainfracloud.org/results/packit/fedora-infra-fedora-messaging-334/fedora-40-x86_64/07105096-fedora-messaging/builder-live.log.gz and here is an excerpt of what causes the failures.

```
INFO: Going to install missing dynamic buildrequires
Updating and loading repositories:
 updates                                100% | 803.1 KiB/s |  23.3 KiB |  00m00s
 fedora                                 100% | 691.1 KiB/s |  20.0 KiB |  00m00s
 Copr repository                        100% |  44.4 KiB/s |   1.5 KiB |  00m00s
Repositories loaded.
Failed to resolve the transaction:
Package "pyproject-rpm-macros-1.12.0-1.fc40.noarch" is already installed.
Package "python3-devel-3.12.2-1.fc40.x86_64" is already installed.
Package "python3-sphinx-1:7.2.6-6.fc40.noarch" is already installed.
Package "python3-packaging-23.2-4.fc40.noarch" is already installed.
Package "python3-pip-23.3.2-1.fc40.noarch" is already installed.
Package "python3-poetry-core-1.8.1-3.fc40.noarch" is already installed.
Package "systemd-rpm-macros-255.3-1.fc40.noarch" is already installed.
Problem: nothing provides requested (python3dist(pytz) < 2024~~ with python3dist(pytz) >= 2023.3)
```

gridhead:
This should do it.

Also as a rule of thumb - please use the simple present verb in your commit messages.

Commit messages describe what a commit does to a repository and not what you did in a commit.

Freedisch:
> Commit messages describe what a commit does to a repository and not what you did in a commit.

Got it,  🙌🏾

abompard:
One more thing: please allow the user to specify the location of the datagrepper instance (defaulting to `apps.fedoraproject.org/datagrepper`) because we may want to replay messages from the staging environment.

Freedisch:
@abompard, I see it makes sense, I'm on  it

Freedisch:
@abompard I have added the support to user to specify a custom URL

fedora_messaging/cli.py
```diff
@@ -452,3 +458,41 @@ def record(exchange, queue_name, routing_key, app_name, limit, file):
...
         exchange, queue_name, routing_key, messages_recorder.collect_message, app_name
     )
+
+
+URL_TEMPLATE = "https://apps.fedoraproject.org/datagrepper/id?id={}&is_raw=true"
+
+
+@cli.command()
+@click.argument("message_id")
+@click.option("--datagrep-url", help=_datagreeper_url, default=URL_TEMPLATE)
```
abompard: Please use the full name: `--datagrepper-url`

fedora_messaging/cli.py
```diff
@@ -91,6 +93,10 @@
     "The name of the exchange to publish to. Can contain ASCII letters, "
     "digits, hyphen, underscore, period, or colon."
 )
+_datagreeper_url = (
```
abompard: There's a typo here, it's `datagrepper` with one `e` and two `p`'s

fedora_messaging/cli.py
```diff
@@ -452,3 +458,41 @@ def record(exchange, queue_name, routing_key, app_name, limit, file):
...
+    """Fetch a message by ID from Datagreeper"""
+    url = datagrep_url.format(message_id)
+    response = requests.get(url, timeout=5)
+    if response.ok:
+        return response.json()
+    else:
+        click.echo(
+            f"Failed to retrieve message from Datagrepper: HTTP {response.status_code}"
+        )
+        return None
```
abompard: Instead of returning `None` you could just let the exception raised by `response.raise_for_status()` propagate upwards and catch it where it's most appropriate to handle it. Otherwise you're doing a test on `response.ok` and another one on `message_data` being `None`. Only one is necessary, this will simplify the code.

fedora_messaging/cli.py
```diff
@@ -452,3 +458,41 @@ def record(exchange, queue_name, routing_key, app_name, limit, file):
...
+    """Replay a message from Datagrepper by its message ID"""
+    message_data = _get_message(message_id, datagrep_url)
+    if message_data:
+        try:
+            _publish_message(message_data)
+            click.echo(f"Message with ID {message_id} has been successfully replayed.")
+        except Exception as e:
+            click.echo(f"Failed to replay message: {e}")
+    else:
+        click.echo(f"Could not find message with ID {message_id}")
```
abompard: Here, it would be more appropriate to raise [click.ClickException](https://click.palletsprojects.com/en/8.1.x/exceptions/), this will instruct click to return with a status code > 0, and would print the error message to stderr.

fedora_messaging/cli.py
```diff
@@ -452,3 +458,41 @@ def record(exchange, queue_name, routing_key, app_name, limit, file):
...
+    if response.ok:
+        return response.json()
+    else:
+        click.echo(
+            f"Failed to retrieve message from Datagrepper: HTTP {response.status_code}"
+        )
+        return None
+
+
+def _publish_message(message_data):
```
abompard: I'm not sure it's very useful to have a separate method just for these two lines.

abompard (changes requested):
This is very good overall, thanks! A few comments though.

fedora_messaging/cli.py
Freedisch: I think I could remove the last function, and integrate it into `replay` method

fedora_messaging/cli.py
```diff
@@ -91,6 +93,10 @@
     "The name of the exchange to publish to. Can contain ASCII letters, "
     "digits, hyphen, underscore, period, or colon."
 )
+_datagrepper_url = (
```
abompard: Please use a `_help` suffix for this variable name like the others

fedora_messaging/cli.py
```diff
@@ -452,3 +458,34 @@ def record(exchange, queue_name, routing_key, app_name, limit, file):
...
+
+def _get_message(message_id, datagrep_url):
+    """Fetch a message by ID from Datagreeper"""
+    url = datagrep_url.format(message_id)
+    try:
+        response = requests.get(url, timeout=5)
+        response.raise_for_status()
+        return response.json()
+    except requests.HTTPError as e:
+        click.ClickException(f"Failed to retrieve message from Datagrepper: {e}")
```
abompard: You have to raise this exception, here you're just instantiating it and raising the original error.

fedora_messaging/cli.py
```diff
@@ -452,3 +458,34 @@ def record(exchange, queue_name, routing_key, app_name, limit, file):
...
+@click.option("--datagrepper-url", help=_datagrepper_url, default=URL_TEMPLATE)
+def replay(message_id, datagrep_url):
+    """Replay a message from Datagrepper by its message ID"""
+    message_data = _get_message(message_id, datagrep_url)
+    if message_data:
+        try:
+            api.publish(Message.load_message(message_data))
+            click.echo(f"Message with ID {message_id} has been successfully replayed.")
+        except Exception as e:
+            click.ClickException(f"Failed to replay message: {e}")
```
abompard: you have to raise this exception

fedora_messaging/cli.py
```diff
@@ -452,3 +458,34 @@ def record(exchange, queue_name, routing_key, app_name, limit, file):
...
+    else:
+        click.ClickException(f"Could not find message with ID {message_id}")
```
abompard: Same here.

fedora_messaging/cli.py
```diff
@@ -452,3 +458,34 @@ def record(exchange, queue_name, routing_key, app_name, limit, file):
...
     )
+
+
+URL_TEMPLATE = "https://apps.fedoraproject.org/datagrepper/id?id={}&is_raw=true"
+
+
+@cli.command()
+@click.argument("message_id")
+@click.option("--datagrepper-url", help=_datagrepper_url, default=URL_TEMPLATE)
+def replay(message_id, datagrep_url):
```
abompard: Hmm, does this actually work? I'm surprised that click is able to match the `--datagrepper-url` option name to the `datagrep_url` variable name. If not I would recommend renaming the variable.

abompard (changes requested):
Thanks for the changes. Besides those fixes, it would be great if you could add unit tests for this new code.

fedora_messaging/cli.py
Freedisch: Actually no, I forgot to update it, I'm doing rn

fedora_messaging/cli.py
```diff
@@ -452,3 +458,33 @@ def record(exchange, queue_name, routing_key, app_name, limit, file):
...
         exchange, queue_name, routing_key, messages_recorder.collect_message, app_name
     )
+
+
+URL_TEMPLATE = "https://apps.fedoraproject.org/datagrepper/id?id={}&is_raw=true"
+
+
+@cli.command()
+@click.argument("message_id")
+@click.option("--datagrepper_url", help=_datagrepper_help, default=URL_TEMPLATE)
```
abompard: Oh yeah one minor thing: the convention is that option words are separated by dashes, not underscores. So it should be `--datagrepper-url`. Click will convert that to underscores for the variable name.

fedora_messaging/cli.py
```diff
@@ -452,3 +458,33 @@ def record(exchange, queue_name, routing_key, app_name, limit, file):
...
+
+
+def _get_message(message_id, datagrepper_url):
+    """Fetch a message by ID from Datagreeper"""
+    url = datagrepper_url.format(message_id)
+    try:
+        response = requests.get(url, timeout=5)
+        response.raise_for_status()
+        return response.json()
+    except requests.HTTPError as e:
```
abompard: It's a minor thing (again) but since you're pretty much putting the entire function inside a `try/except` block, I would just put that block in the calling function instead, something like:

```py
def replay(message_id, datagrepper_url):
    try:
        message_data = _get_message(message_id, datagrepper_url)
    except requests.HTTPError as e:
        raise ...

**Review by abompard — CHANGES_REQUESTED** [review]
Another minor thing. And please don't forget the unit tests! I can help with that if you're unfamiliar with unit testing.

File: fedora_messaging/cli.py (hunk 10 above)
Comment (abompard) [reply fedora_messaging/cli.py]: No worries, but a unit test would have caught that ;-)

**Comment by Freedisch** [comment]
> Another minor thing. And please don't forget the unit tests! I can help with that if you're unfamiliar with unit testing.

I'm familiar with unit tests, I will check the existing ones and follow the same workflow.

File: fedora_messaging/cli.py (hunk 12 above)
Comment (Freedisch) [reply fedora_messaging/cli.py]: I was wondering what if I remove the `try/except` from the `_get_message` function and wrap all these operations in a single try/except block that catches and handles all exceptions? will this be much cleaner?
```python
 try:
        message_data = _get_message(message_id, datagrepper_url)
        api.publish(message.load_message(message_data))
        click.echo(f"Message with ID {message_id} has been successfully replayed.")
 except requests.HTTPError:
        raise click.ClickException(f"Could not find message with ID {message_id}")
 except Exception as e:
        raise click.ClickException(f"Failed to replay message: {e}")
```

Freedisch:
Test cases results:

Screenshot from 2024-03-15 20-50-05

Screenshot from 2024-03-15 20-49-03

fedora_messaging/cli.py
abompard: Very good question.
I'd prefer keeping the try/except blocks specific to what we want to handle, for example we know how to handle `HTTPError`s for the `_get_message()` call, so we handle it, but we may not want to catch the generic exception there. So, multiple blocks seems cleaner to me.

abompard (changes requested):
Please make the last try/except change and it should be all good :-)

fedora_messaging/cli.py
```diff
@@ -452,3 +458,28 @@ def record(exchange, queue_name, routing_key, app_name, limit, file):
...
+@click.argument("message_id")
+@click.option("--datagrepper_url", help=_datagrepper_help, default=URL_TEMPLATE)
+def replay(message_id, datagrepper_url):
+    """Replay a message from Datagrepper by its message ID"""
+    try:
+        message_data = _get_message(message_id, datagrepper_url)
+        if message_data:
+            api.publish(message.load_message(message_data))
+            click.echo(f"Message with ID {message_id} has been successfully replayed.")
+    except requests.HTTPError as e:
```
abompard, line 473: Actually this exception can only happen during the `_get_message()` call, so it should not contain the rest (the `if` block).

abompard (changes requested):

abompard:
OK great! Only the `--datagrepper-url` option name to fix! :-)

Freedisch:
> OK great! Only the `--datagrepper-url` option name to fix! :-)

Actually, I thought I fixed it already 😅

abompard (approved):
Looks great, thanks for your work and your attention to detail.
I'll merge it and release a new version soon.

abompard:
Ah, yeah, the unit test needs to be adjusted as well.

abompard:
OK thanks! Could you rebase and solve the conflicts? If you have any issues doing that please ask.

Freedisch:
> OK thanks! Could you rebase and solve the conflicts? If you have any issues doing that please ask.

I was wondering also If I should squash all these commits in one?

abompard:
> I was wondering also If I should squash all these commits in one?

Yeah that would be preferable.

abompard:
Merged, thanks again for your contribution.

Issue #303, opened 2023-09-11
Add an `async` function to publish messages


abompard:
Add an asyncio-based publish function.

Freedisch:
I assume you are referring to the publish function present in the API. I will check it and run tests to ensure nothing breaks after the changes.
```python
def publish(message, exchange=None, timeout=30):
    """
    Publish a message to an exchange.
```

abompard:
This is actually very far from a simple issue, I wouldn't start with this one.

bajajharsha:
Hi, I'd like to work on this issue

abompard:
Please see the comment above @bajajharsha .

bajajharsha:
@abompard Can you guide me on which one is comparatively an easier issue that I can start with?

nwekealex65:
@abompard I do like to work on this, I have experience working with asyncio

PR #336, opened 2024-03-12
feat: Add asyncio support to publish function

nwekealex65:
This pull request addresses issue #303 

Changes Made:
- Added support for asynchronous programming using asyncio to the `publish function`.
- Utitlized async await syntax to make the function asynchronous.
- Introduced error handling for asyncio.TimeoutError to handle timeouts.

Related Issue:
Issue #303 

Please review and provide feedback. Thank you!

### DIFF
--- a/fedora_messaging/api.py
+++ b/fedora_messaging/api.py
@@ -3,6 +3,7 @@
 
 import inspect
 import logging
+import asyncio
 
 import crochet
 from twisted.internet import defer, reactor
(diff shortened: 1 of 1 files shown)


abompard:
The problem that needs fixing is that we want to allow asyncio applications to publish messages within the asyncio framework, not needing the thread pool. So what we need is an async publish function, not a sync publish function that calls `asyncio.run()`. This is why it's harder, we need to find a way to integrate the async function in Twisted.

nwekealex65:
> The problem that needs fixing is that we want to allow asyncio applications to publish messages within the asyncio framework, not needing the thread pool. So what we need is an async publish function, not a sync publish function that calls `asyncio.run()`. This is why it's harder, we need to find a way to integrate the async function in Twisted.

So how do we go about this? Any suggestion?

abompard:
Well I have started working on it a couple months ago, spent a few days on it and stopped because I had other things to do and it was taking too long. I would suggest finding another issue, as I said before this one is definitely not a simple issue. If you manage do to that before the end of March, you should mentor me ;-)

Issue #229, opened 2020-10-13
docs: document how to consume messages with a message schema


pypingou:
The current docs has examples and instructions on how to send messages using a message schema but not of consuming messages that have a schema.

How does a consumer know that a message it received follows a schema? How does it convert the base Message into a more specific object?

jeremycline:
All the standard Python object introspection interfaces should be sufficient. The consumer is called with an instance of the class that defines the schema, if it's installed. If it's not, it'll get a plain Message. Anything from ``isinstance`` to poking around in the attached jsonschema (although if you're doing this you probably shouldn't be) should work as you'd expect. I think it's fine to add a note to the consumer docs saying something like that.

anshikavashistha:
@jeremycline @pypingou May I work on this issue?

PR #337, opened 2024-03-13
doc: Update usage.rst

nwekealex65:
Added documentation for how to consume messages with a message schema. This addresses #229. It explains how to consume message schema

### DIFF
--- a/docs/tutorial/usage.rst
+++ b/docs/tutorial/usage.rst
@@ -45,6 +45,43 @@ interface, you'll see that a message has been sent to the ``amq.topic``
 exchange. However, since noone is listening to this topic, the message has been
 discarded. Now, we'll setup listeners.
 
+
+
+Consuming Messages
+------------------
+
+Consuming messages with a message schema ensures that messages adhere to predefined standards, facilitating interoperability and consistency within the Fedora Messaging ecosystem. This section provides guidance on how consumers can validate received messages against a schema and convert the base `Message` into more specific objects based on the schema definition.
+
+Validating Messages Against Schema
+----------------------------------
+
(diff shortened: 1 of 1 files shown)


docs/tutorial/usage.rst
```diff
@@ -45,6 +45,43 @@ interface, you'll see that a message has been sent to the ``amq.topic``
...
+
+Consuming Messages
+------------------
+
+Consuming messages with a message schema ensures that messages adhere to predefined standards, facilitating interoperability and consistency within the Fedora Messaging ecosystem. This section provides guidance on how consumers can validate received messages against a schema and convert the base `Message` into more specific objects based on the schema definition.
+
+Validating Messages Against Schema
+----------------------------------
+
+Before processing a received message, consumers must validate it against the associated message schema. This ensures that the message structure conforms to the expected format defined in the schema. The `jsonschema` library is commonly used for this purpose within the Fedora Messaging project.
```
abompard: This is actually not true, the message is already validated as part of the consuming process (in `message.get_message()`.

abompard (changes requested):

docs/tutorial/usage.rst
nwekealex65: So I should remove this part? any suggestions?

abompard: Yeah, but that means removing everything actually, because the introduction in "Consuming Messages" is wrong too: it mentions how consumers can validate received messages, and they don't need to.

abompard:
This isn't really what Jeremy mentions in the ticket comment.

nwekealex65:
@abompard I made some changes to this, can review please?

PR #350, opened 2024-03-31
Improving 'id' schema documentation

### LINKED ISSUE #220: Some schema "id" field is not documented
We should have some slight note in docs about `"id": "http://fedoraproject.org/message-schema/mailman#",`.   E.g. should we keep this field byte-by-byte the same for all the messages?  Should each message have it's own item?

vidit-maheshwari:
## Summary:

Fixes #220
This PR updates the documentation to provide clarity on the usage of the `id` (or `$id`) field in message schemas within Fedora Messaging. 

## Description:

This PR updates the documentation to provide clarity on the usage of the `id` (or `$id`) field in message schemas within Fedora Messaging. The `id` field serves as a URI identifier for each schema, distinguishing between different schema packages. It is important to maintain consistency within a specific schema package while ensuring uniqueness across packages.

### Changes Made:

Updated the documentation to emphasize the purpose and significance of the `id` field in message schemas.

### DIFF
--- a/docs/tutorial/schemas.rst
+++ b/docs/tutorial/schemas.rst
@@ -57,8 +57,10 @@ Schema format
 This dictionary describes the possible keys and types in the JSON object being
 validated, using the following reserved keys:
 
-- ``id`` (or ``$id``): an URI identifing this schema. Change the last part of
-  the example URL to use your app's name.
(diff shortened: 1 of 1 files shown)


gridhead:
@praiskup, as the original poster for #220, could you please verify if this PR fits the bill?

PR #359, opened 2024-04-17
Add a new reconsume command to the CLI

jeremycline:
This command is similar to the "replay" command, except it initializes a
consumer and passes it the message retrieved from datagrepper.

I recently implemented a new consumer and wrote myself a little CLI
which did this in order to test my consumer from an existing message.
It'd be nice for it to be available by default.

I've marked this as a draft for now since I still need to write tests and I'm not in love with the command name. Any alternate names definitely welcome.


abompard (approved):
Looks good, thanks!

PR #375, opened 2024-06-21
Add back api._twisted_publish with deprecation warning

mattiaverga:
`api._twisted_publish` is (was) used for mocking tests.
The removal of that method in 3.6.0 without previous warnings broke Bodhi tests (and I think other dependent packages too). The removal should be handled by provide a proper warning when the method is used and wait a decent amount of time before complete removal.

This PR adds back the deprecated method, which just emits a deprecation notice and routes to the new method, which seems 1:1 compatible.
I just don't know if the decorators should be applied to the deprecated method as well.

### DIFF
--- a/fedora_messaging/api.py
+++ b/fedora_messaging/api.py
@@ -2,6 +2,7 @@
 
 import inspect
 import logging
+import warnings
 
 import crochet
 from twisted.internet import defer, reactor, threads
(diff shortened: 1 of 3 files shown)


abompard:
Ah, sorry about the breakage, I assumed that functions starting with an underscore would be private and people woudn't use them directly, but it's true that it can easily happen when mocking.
The `fedora_messaging.testing` module comes with functions and fixtures to help with testing, is there something that you'd need me to add there? I'd rather provide the testing functionality as part of fedora_messaging rather that force people to mock private functions.

mattiaverga:
The reason behind mocking `api._twisted_publish` is to ensure a proper failure message if, for some reason, a message publish was not properly handled during tests.
See https://github.com/fedora-infra/bodhi/commit/3eb32044de764e3ed94bdc5b9ab83253f40cbafd

But I originally created this PR because I was quite sure to have seen an example about mocking that method in fedora-messaging docs (which, however, I cannot find anymore...).

So, I'm unsure if this can be useful to some other places other than bodhi or not. BTW about bodhi I [solved](https://github.com/fedora-infra/bodhi/commit/16710bf52a4c597cac8202cb3eba54ae1278e74c) the transition with a try / except AttributeError to handle both the new and old fedora-messaging versions.

PR #378, opened 2024-07-05
Indicate which package a schema comes from when missing

### LINKED ISSUE #187: Add messages header that points to the Python package for the schema
Publishers using schema add a header to identify the schema, but from the consumer side it's not obvious where to get the schema.

We can use the entry point of the message being published to inspect that package's metadata and provide things like the URL and so on. Then on the consumer side we can log more helpful messages when users are missing the schema.

abompard:
Fixes: #187


LenkaSeg:
LGTM!

PR #385, opened 2024-07-11
Add an embedded HTTP server to monitor the service

### LINKED ISSUE #380: Add a way to be probed/monitored by Openshift
It would be nice if fedora-messaging consumers had a way to support [liveness/readiness checks in OpenShift](https://docs.openshift.com/container-platform/4.16/applications/application-health.html).

Maybe using an embedded HTTP server that would reply to liveness on startup and reply to readiness when the consumer has started.

Bonus: return in the HTTP call the number of messages processed since the startup of the consumer, so that it can be graphed.

The listening address and the port would have to be configurable in the config file.

abompard:
This changeset adds an embedded HTTP server to monitor consumers and producers, and get statistics about activity.

Fixes: #380

### DIFF
--- a/docs/user-guide/consuming.rst
+++ b/docs/user-guide/consuming.rst
@@ -197,6 +197,42 @@ consumers to use if they need configuration options. Refer to the
 :ref:`conf-consumer-config` in the Configuration documentation for details.
 
 
+Monitoring
+==========
+
+A Fedora Messaging consumer can start a dedicated HTTP server to let users
+monitor its state. You can configure the port in the configuration file under
+the ``[monitoring]`` section, with ``address`` and ``port``.
+If the section is empty, the monitoring service will be disabled.
+The default value for ``address`` is an empty string, which means that the
+monitoring server will listen on all interfaces.
+There is no default value for ``port``, you will have to choose a port.
+
+When a consumer is running with the monitoring server enabled, you can get the
+following data::
+
+    $ curl http://localhost:8070/live
+    {"status": "OK"}
+
+    $ curl http://localhost:8070/ready
+    {"consuming": true, "published": 0, "consumed": {"received": 0, "processed": 0, "dropped": 0, "rejected": 0, "failed": 0}}
+
+The ``/live`` endpoint always returns the same JSON data.
+The statistics in the ``/ready`` endpoint are gathered from the start of the
+service. They mean:
+
+- ``consuming``: whether the consumer is running or not.
+- ``published``: amount of messages published on the bus.
+- ``consumed.received``: amount of messages received from the bus.
+- ``consumer.processed``: amount of messages successfully processed by the consumer
+- ``consumer.dropped``: amount of messages dropped by the consumer (by raising the
+  :class:`fedora_messaging.exceptions.Drop` exception).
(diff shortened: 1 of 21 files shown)


Conan-Kudo (approved):

docs/user-guide/consuming.rst
```diff
@@ -197,6 +197,40 @@ consumers to use if they need configuration options. Refer to the
...
 
 
+Monitoring
+==========
+
+A Fedora Messaging consumer can start a dedicated HTTP server to let users
+monitor its state. You can configure the port in the configuration file under
+the ``[monitoring]`` section, with ``address`` and ``port``.
+The default address is empty, which means that the monitoring server will
+listen on all interfaces. The default port is ``0``, which means that the
```
jeremycline: This is a slightly surprising behavior since 0 usually means "bind to a free port" (try, for example, `nc -l ::1 0` and then see the port with `ss -plnt`).

Maybe just have the feature off if the section doesn't exist in the configuration and require both values to be set (or pick static defaults) if the section does exit.

docs/user-guide/consuming.rst
```diff
@@ -197,6 +197,40 @@ consumers to use if they need configuration options. Refer to the
...
+monitoring server is disabled.
+
+When a consumer is running with the monitoring server enabled, you can get the
+following data::
+
+    $ curl http://localhost:8070/live
+    {"status": "OK"}
```
jeremycline, line 215: It'd be nice to include a bit more documentation on this endpoint to indicate what set of values it returns. Looking at the code it seems it only ever returns this, so just calling that out explicitly in the docs would be good.

fedora_messaging/config.py
```diff
@@ -287,6 +287,20 @@ def callback(message):
...
+
+monitoring
+----------
+The options for the embedded HTTP server dedicated to monitoring the service.
+This is where you can configure the address and the port to be listened on.
+The defaults are::
+
+    {
+        'address': '',  # listen on all interfaces
+        'port': 8070,
```
jeremycline: This doesn't line up with the documentation elsewhere and the code

jeremycline (changes requested):
I like the feature - a couple notes on the docs and defaults, nothing major though.

tests/integration/test_cli.py
```diff
@@ -99,3 +118,90 @@ def test_consume_halt_with_exitcode(callback, exit_code, msg, queue, cli_conf):
...
+    # Check stats
+    response = yield http_client.request(b"GET", base_url + b"/ready")
+    body = yield readBody(response)
+    assert body == (
+        b'{"consuming": true, "published": 0, "consumed": {"received": 2, "processed": 1, "dropped": 1, '
+        b'"rejected": 0, "failed": 0}}\n'
+    )
+
+    # Don't check the reject action or the message will be put back in the queue and loop forever
+    # Don't check a generic processing failure because it stops the consumer instantly and we can't look
```
jeremycline, line 193: I guess it could log the final stats on exit, if available, and then you could inspect the log, but whether you bother with all that is totally up to you.

jeremycline (approved):

PR #477, opened 2025-06-11
Add typing to the entire library, and check it with mypy and pyright

abompard:
This changeset adds typing to the entire library.
It also drops support for Python 3.8 and Pika < 1.0


abompard:
Merged in 7e0f19b8

PR #532, opened 2025-11-04
Install a new event loop explicitly

### LINKED ISSUE #531: fedora-messaging broken with Python 3.14
I've just tried to upgrade bodhi in staging to F43 and noticed that fedora-messaging seems completely broken with Python 3.14. bodhi-consumer pod fails with:
```
Traceback (most recent call last):
File "/usr/bin/fedora-messaging", line 5, in <module>
from fedora_messaging.cli import cli
File "/usr/lib/python3.14/site-packages/fedora_messaging/cli.py", line 26, in <module>
asyncioreactor.install()
~~~~~~~~~~~~~~~~~~~~~~^^
File "/usr/lib/python3.14/site-packages/twisted/internet/asyncioreactor.py", line 304, in install
reactor = AsyncioSelectorReactor(eventloop)
File "/usr/lib/python3.14/site-packages/twisted/internet/asyncioreactor.py", line 50, in __init__
_eventloop: AbstractEventLoop = get_event_loop()
~~~~~~~~~~~~~~^^
File "/usr/lib64/python3.14/asyncio/events.py", line 715, in get_event_loop
raise RuntimeError('There is no current event loop in thread %r.'
% threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'MainThread'.
```

I've also noticed that in koji builds no tests are performed.

jeremycline:
Python introduced some breaking changes[0] which cause it to no longer implicitly create an event loop. Create a new one and install it explicitly so Twisted doesn't collapse at import.

[0] https://docs.python.org/3/whatsnew/3.14.html#id10

Fixes: #531

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -71,6 +71,7 @@ jobs:
           - py311-unittest
           - py312-unittest
           - py313-unittest
+          - py314-unittest
 
   integration-tests:
     name: Integration tests
@@ -106,6 +107,7 @@ jobs:
           - py311-integration
           - py312-integration
           - py313-integration
+          - py314-integration
 
 
   # https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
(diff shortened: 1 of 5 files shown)


jeremycline:
There's not an obvious way to get ahead of pytest initializing pytest_twisted which immediately falls over. I guess possibly some wrapper that sets things up before running pytest, but I'm not inclined to crawl into that particular rabbit hole.

PR #558, opened 2025-12-16
refactor(twisted/consumer): update asyncio.iscoroutinefunction to inspect.iscoroutine

### LINKED ISSUE #546: Deprecation warnings: asyncio.iscoroutinefunction
On upgrading our openQA deployments to Fedora 43 I noticed some deprecation warnings for fedora-messaging, I figured I'd report them so we can hopefully fix them before they become errors.

There are two deprecation warnings relating to asyncio.iscoroutinefunction:
```
/usr/lib/python3.14/site-packages/fedora_messaging/twisted/consumer.py:51: DeprecationWarning: 'asyncio.iscoroutinefunction' is deprecated and slated for removal in Python 3.16; use inspect.iscoroutinefunction() instead
/usr/lib/python3.14/site-packages/fedora_messaging/twisted/consumer.py:52: DeprecationWarning: 'asyncio.iscoroutinefunction' is deprecated and slated for removal in Python 3.16; use inspect.iscoroutinefunction() instead
```

servusdei2018:
Fixes #546 

[`asyncio.iscoroutinefunction` is deprecated](https://github.com/python/cpython/pull/122875) as of Python 3.14 (stable) and will be removed in Python 3.16.

This PR changes those deprecated references to use [`inspect.iscoroutinefunction`](https://docs.python.org/3/library/inspect.html#inspect.iscoroutinefunction), added in Python 3.5.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2802,30 +2802,30 @@ files = [
 
 [[package]]
 name = "ruff"
-version = "0.16.3"
+version = "0.16.4"
 description = "An extremely fast Python linter and code formatter, written in Rust."
 optional = false
 python-versions = ">=3.7"
 groups = ["dev"]
 files = [
-    {file = "ruff-0.16.3-py3-none-linux_armv6l.whl", hash = "sha256:0c5710e247a58a4521e66e124ba9a74655b414f61ba3a2e9e3811e11098f48f7"},
-    {file = "ruff-0.16.3-py3-none-macosx_10_12_x86_64.whl", hash = "sha256:fe155130631a2471fd2e14a7a664a4dfbd7194b8229c3d7b2a40b21178639081"},
-    {file = "ruff-0.16.3-py3-none-macosx_11_0_arm64.whl", hash = "sha256:e2ed719e14aa64d895c2ee922594a90a43c861a93f0575a95ff8c47cdbd13eb9"},
-    {file = "ruff-0.16.3-py3-none-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:9e0b1da805eb043654645d74d5de1e5ce2edc686e40790d2b86f56d71cc06a84"},
-    {file = "ruff-0.16.3-py3-none-manylinux_2_17_armv7l.manylinux2014_armv7l.whl", hash = "sha256:a37bdea0bbe21780f590bf437d6412c8c4e1b6cd010f91a65c2c40c5e5f5f870"},
-    {file = "ruff-0.16.3-py3-none-manylinux_2_17_i686.manylinux2014_i686.whl", hash = "sha256:09571e6d1288ed9be475207a3ac04ada404f1cd898104be0f6ab8d7df438575b"},
-    {file = "ruff-0.16.3-py3-none-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl", hash = "sha256:2c18c5a101eb540010638cc1ff3c84944d3adb3df62b8d98ca8f22ba484d3413"},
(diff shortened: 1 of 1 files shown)


AdamWill:
This would also need to change the comment block that states a specific reason for using asyncio. over inspect. And we should verify that we don't need this to work on Python < 3.10, which isn't a safe assumption - RHEL 9's system Python is 3.9, so if we run anything that uses fedora-messaging on EL 9, it could potentially be an issue.

servusdei2018:
> This would also need to change the comment block that states a specific reason for using asyncio. over inspect. And we should verify that we don't need this to work on Python < 3.10, which isn't a safe assumption - RHEL 9's system Python is 3.9, so if we run anything that uses fedora-messaging on EL 9, it could potentially be an issue.

@AdamWill Just added backwards compatibility for Python pre-3.10.

AdamWill:
So I looked into the cpython commit history, and it seems the change to make the `inspect` version match the `asyncio` version actually landed in 3.10.6:
```
[adamw@toolbx fedora-toolbox-44 cpython ((v3.10.4))]$ git checkout v3.10.5
Previous HEAD position was 9d38120e335 Python 3.10.4
HEAD is now at f3771539678 Python 3.10.5
[adamw@toolbx fedora-toolbox-44 cpython ((v3.10.5))]$ git log --oneline Lib/inspect.py | grep gh-84753
[adamw@toolbx fedora-toolbox-44 cpython ((v3.10.5))]$ git checkout v3.10.6
Previous HEAD position was f3771539678 Python 3.10.5
HEAD is now at 9c7b4bd1646 Python 3.10.6
[adamw@toolbx fedora-toolbox-44 cpython ((v3.10.6))]$ git log --oneline Lib/inspect.py | grep gh-84753
47f23b2d8a1 gh-84753: Make inspect.iscoroutinefunction() work with AsyncMock (GH-94050) (GH-94461)
```
so I think we should tweak this to:
```
if sys.version_info < (3, 10, 6):
```
aside from that I think it's probably good to go.

AdamWill:
That seems...overcomplicated, now?

Why not just do this?

```
--- a/fedora_messaging/twisted/consumer.py
+++ b/fedora_messaging/twisted/consumer.py
@@ -4,7 +4,12 @@
 
 import asyncio
 import logging
+import sys
 import uuid
+if sys.version_info < (3, 10, 6):
+    from asyncio import iscoroutinefunction
+else:
+    from inspect import iscoroutinefunction
 from collections.abc import Coroutine, Generator
 from typing import Any, Callable, cast, Optional, TYPE_CHECKING, Union
 
@@ -44,11 +49,8 @@ def is_coro(
     func_or_obj: config.CallbackType,
 ) -> "TypeGuard[Callable[..., Coroutine[Any, Any, Any]]]":
     """Tests if a function is a coroutine function or a callable coroutine object."""
-    # Until Python 3.10, inspect.iscoroutinefunction() will fail to identify AsyncMocks
-    # as coroutines: https://github.com/python/cpython/issues/84753
-    # Use asyncio.iscoroutinefunction() instead.
-    return asyncio.iscoroutinefunction(func_or_obj) or (
-        callable(func_or_obj) and asyncio.iscoroutinefunction(func_or_obj.__call__)  # type: ignore
+    return iscoroutinefunction(func_or_obj) or (
+        callable(func_or_obj) and iscoroutinefunction(func_or_obj.__call__)  # type: ignore
     )
```

AdamWill:
I think this is probably fine now except it doesn't need to be 12 commits :D It could be squashed to one.

I'm not sure who has permissions to merge in this repo. @abompard ?

jeremycline:
I have no idea why GitHub closed this, all I did was rebase and squash it. And it doesn't seem like I can re-open it either...

PR #582, opened 2026-06-23
Handle BadDeclaration when setting up the bindings as well

abompard:
This commit moves the `BadDeclaration` handling to an errback.

There's also some refactoring in the `test_api.py` file to factor out some common operations and make things a bit clearer (I think).

Check it out when you have time please @jeremycline


jeremycline (approved):
Nice cleanups in the tests, and the errback is much more consistent. Thanks for tidying that up!

